PR #2, opened 2015-09-02
Tickets/dm 3566

n8pease:
Adds the butler datasetType feature and supporting tests.
If adding the data files to this project is not the best approach let's discuss other ways to do it.


python/lsst/daf/persistence/butler.py
```diff
@@ -136,8 +138,15 @@ def __init__(self, root, mapper=None, **mapperArgs):
         self.log = pexLog.Log(pexLog.Log.getDefaultLog(),
                 "daf.persistence.butler")
 
-    def getKeys(self, datasetType=None, level=None):
+    def registerDatasetTypeAlias(self, type, alias):
```
ktlim: Use of `type` is discouraged as it is a Python built-in.  I tend to use `datasetType` instead.

python/lsst/daf/persistence/butler.py
```diff
@@ -433,8 +451,18 @@ def _read(self, pythonType, location):
         return results
 
     def __reduce__(self):
-        return (_unreduce, (self.mapper,))
+        return (_unreduce, (self.mapper, self.datasetTypeAliasDict))
+
+    def __resolveDatasetTypeAlias(self, type):
```
ktlim: You should generally use a single underscore to indicate a "private/protected" method, not a double underscore (which, among other things, prevents overriding in a subclass).

python/lsst/daf/persistence/butler.py
```diff
@@ -433,8 +451,18 @@ def _read(self, pythonType, location):
...
+        resolvedType = type
```
ktlim: I think this whole method could be replaced by `return self.datasetTypeAliasDict.get(datasetType, datasetType)`.  But it doesn't quite do enough.  First of all, the specification is to replace "`@alias`" (not just "`alias`") with "`datasetType`".  Second, the specification is also to replace "`@alias`" when it occurs as a *substring* of the original type: "`@coadd`" -> "`deepStack`" also implies "`@coadd_src`" -> "`deepStack_src`".

n8pease:
Fritz messed up this PR -- starting a new one

PR #3, opened 2015-09-02
Implements butler datasetType aliases and related unit tests

n8pease:
After instantiating a butler, you can call registerDatasetTypeAlias
to use a different name for standard dataset types, or change the
current dataset that is being operated with a single change (e.g.
from raw to calexp).
All storage is still done using the 'real' datasetType name (not
the alias type).


python/lsst/daf/persistence/butler.py
```diff
@@ -136,8 +139,36 @@ def __init__(self, root, mapper=None, **mapperArgs):
...
+                           first character.
+        @param datasetType (str) the keyword that will be used when @alias is passed into datasetType. it may
+                                 not contain '@'
+        """
+
+        #verify formatting of alias:
+        # it can have '@' as the first character (if not it's okay, we will add it) or not at all.
+        atLoc = alias.rfind('@')
+        if atLoc is -1:
+            alias = "@%s" %(alias)
```
ktlim: I think `"@" + alias` might be simpler here.

python/lsst/daf/persistence/butler.py
```diff
@@ -136,8 +139,36 @@ def __init__(self, root, mapper=None, **mapperArgs):
...
+        elif atLoc > 0:
+            raise RuntimeError("Badly formatted alias string:%s" %(alias))
```
ktlim: I'd put a space after the colon.  It's slightly more accurate to use a comma after `alias` to indicate that this is a one-item tuple.  But using `.format()` instead of `%` might be even better.

python/lsst/daf/persistence/butler.py
```diff
@@ -136,8 +139,36 @@ def __init__(self, root, mapper=None, **mapperArgs):
...
+
+        # verify that datasetType does not contain '@'
+        if datasetType.count('@') != 0:
+            raise RuntimeError("Badly formatted type string:%s" %(datasetType))
+
+        # verify that the alias keyword does not start with another alias keyword,
+        # and vice versa
+        for key in self.datasetTypeAliasDict:
+            if key.find(alias) is 0 or alias.find(key) is 0:
```
ktlim: I think `.startswith()` is better than `.find()` here.

tests/butlerAlias.py
```diff
@@ -0,0 +1,130 @@
...
+        keys = self.butler.getKeys(self.datasetType)
+        expectedKeyValues = {'filter':['g', 'r'], 'visit':[1, 2, 3]}
+        for key in keys:
+            val = self.butler.queryMetadata(self.datasetType, key)
+            self.assertEqual(val, expectedKeyValues[key])
+
+    def testDatasetExists(self):
+        # I was expecting to be able to iterate over the files as they are named and see
+        # that the datasets exist in the files as they are named (for example, a file named
+        # raw_v1_fg.fits.gz contains visit 1 and filters f & g. However, the test failed and
```
ktlim: `raw_v1_fg.fits.gz` contains only filter `g`; the `f` stands for "filter".

tests/butlerSubset.py
```diff
@@ -83,9 +83,20 @@ def __init__(self):
 class ButlerSubsetTestCase(unittest.TestCase):
     """A test case for the subset capability of the data butler."""
 
+    # create some random alias typenames
+    calexpTypeName = "@goBears"
```
ktlim: :)

tests/butlerAlias.py
```diff
@@ -0,0 +1,130 @@
...
+        self.assertEqual(self.butler.datasetExists(self.datasetType, {'filter':'g', 'visit':1}), True)
+        self.assertEqual(self.butler.datasetExists(self.datasetType, {'filter':'g', 'visit':2}), True)
+        self.assertEqual(self.butler.datasetExists(self.datasetType, {'filter':'r', 'visit':3}), True)
+        
+        # test a few values that are expected to be false:
+        self.assertEqual(self.butler.datasetExists(self.datasetType, {'filter':'f', 'visit':1}), False)
+        self.assertEqual(self.butler.datasetExists(self.datasetType, {'filter':'r', 'visit':1}), False)
+        self.assertEqual(self.butler.datasetExists(self.datasetType, {'filter':'g', 'visit':3}), False)
+
+    def testGet(self):
```
ktlim: This overrides and replaces the former test with the same name...

python/lsst/daf/persistence/butler.py
n8pease: Admittedly it's a little more verbose, but I've been burned many times when the RHS var is an int and the execution raises because you can't concat str and int vars.

tests/butlerAlias.py
n8pease: yeah, I figured that out eventually. I'll fix the comment.

python/lsst/daf/persistence/butler.py
PaulPrice: If that's what you're worried about, cast it explicitly:

    alias = "@" + str(alias)

PR #5, opened 2015-10-29
U/lauren/dm 3911

n8pease:


python/lsst/daf/persistence/butler.py
```diff
@@ -34,6 +34,64 @@
...
+    outDir, outName = os.path.split(name)
+    safeMakeDir(outDir)
+    temp = tempfile.NamedTemporaryFile(dir=outDir, prefix=outName, delete=False)
+    tempName = temp.name
+    temp.close() # We don't use the fd, just want a filename
+    try:
+        yield tempName
+    finally:
+        os.rename(tempName, name)
+        setFileMode(name)
```
n8pease: safeMakeDir, safeFileMode, SafeFile, and SafeFilename should go in a separate file. It should be included by butler.py, it does not needed to be added to the includes in __init__.py

n8pease:
just FYI: I was looking at this with KT and he pointed out a couple very-edge-cases that these fixes don't cover. I created DM-4215 to cover those and will work on those at some point.

python/lsst/daf/persistence/butler.py
laurenam: Sure.  I will put these functions into a separate file named safeFileIo.py and import from it in butler.py.

PR #6, opened 2015-11-13
butler-specific policy class

n8pease:
adds a pure python butler policy class. basically nested dicts,
but implemented as a class that adds some functionality to
support policy-ish api, such as dot delimited hierarchical
access.

### DIFF
--- a/python/lsst/daf/persistence/policy.py
+++ b/python/lsst/daf/persistence/policy.py
@@ -1,0 +1,387 @@
+#!/usr/bin/env python
+
+#
+# LSST Data Management System
+# Copyright 2015 LSST Corporation.
+#
+# This product includes software developed by the
+# LSST Project (http://www.lsst.org/).
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+
+import collections
+import copy
+import os
+import UserDict
+import warnings
+import yaml
+
+import lsst.pex.policy as pexPolicy
+import lsst.utils
+
+class Policy(UserDict.UserDict):
+    """Policy implements a datatype that is used by Butler for configuration parameters.
+    It is essentially a dict with key/value pairs, including nested dicts (as values). In fact, it can be
+    initialized with a dict. The only caveat is that keys may NOT contain dots ('.'). This is explained next:
+    Policy extends the dict api so that hierarchical values may be accessed with dot-delimited notiation.
+    That is, foo.getValue('a.b.c') is the same as foo['a']['b']['c'] is the same as foo['a.b.c'], and either
+    of these syntaxes may be used.
+
+    Storage formats supported:
+    - yaml: read and write is supported.
+    - pex policy: read is supported, although this is deprecated and will at some point be removed.
+    """
+
+    defaultPolicyInitData = collections.namedtuple('defaultPolicyInitData', 'productName filePath repos')
+    # Contains data for loading a default policy from a package:
+    # product name: the name of the pacage
+    # filePath: the name of the policy file
+    # the file path from the root of the directory to the dir that contains the file.
+
+    def __init__(self, other=None, preference=('data', 'policy', 'pexPolicy', 'filePath', 'defaultInitData'),
+                 **kwargs):
+        """Constructor. Accepts an optional vars, one of which will be used to initalize the Policy.
+        If more than initializer var is populated, one will be selected in order specified by 'preference'.
+
+        :param preference: a tuple in order of preference to be used if more than 1 initialzer var is
+                           populated.
+        :param kwargs:
+            data: a dict with initial key-value pairs.
+            policy: an instance of a Policy that should be copied into this Policy.
+            filePath: path to a policy file. May be .yaml or .paf
+            pexPolicy: an instance of a pex policy to copy into this Policy
+            defaultInitData: policy data with a populated instance of a defaultInitData
+                             e.g. Policy.defaultInitData('daf_persistence', 'testPolicy.yaml', 'tests')
+
+        :return:
+        """
+        UserDict.UserDict.__init__(self)
+        self.__init(other, preference, **kwargs)
+
+    def __init(self, other, preference, **kwargs):
+
+        for p in preference:
+            if p in kwargs:
+                if p == 'policy':
+                    self.data = {}
+                    self.merge(kwargs['policy'])
+                    return
+                if p == 'filePath':
+                    self.__initFromFile(kwargs['filePath'])
+                    return
+                if p == 'pexPolicy':
+                    self.__initFromPexPolicy(kwargs['pexPolicy'])
+                    return
+                if p == 'defaultInitData':
+                    self.__initFromDefaultPolicy(preference, *kwargs['defaultInitData'])
+                    return
+                if p == 'data':
+                    # don't assign other to self.data directly; need assign via parse in self.update.
+                    self.update(kwargs['data'])
+                    return
+
+        if other is not None:
+            import pdb; pdb.set_trace()
+
+    def __initFromFile(self, paths):
+        """Load a file from path. If path is a list, will pick one to use, according to order specified
+        by extensionPreference.
+
+        :param path: string or list of strings, to a persisted policy file.
+        :param extensionPreference: the order in which to try to open files. Will use the first one that
+        succeeds.
+        :return:
+        """
+        extensionPreference = ('yaml', 'paf')
+        matchedPath = None
+        if isinstance(paths, basestring):
+            paths = [paths]
+        for ext in extensionPreference:
+            if matchedPath is not None:
+                break
+            for path in paths:
+                if path.endswith(ext):
+                    matchedPath = path
+                    break
+        policy = None
+        if matchedPath is None:
+            raise IOError("Could not find policy at %s" %paths)
+        elif matchedPath.endswith('yaml'):
+            self.__initFromYamlFile(matchedPath)
+        elif matchedPath.endswith('paf'):
+            policy = pexPolicy.Policy.createPolicy(matchedPath)
+            self.__initFromPexPolicy(policy)
+        else:
+            raise RuntimeError("Unhandled policy file type:%s" %matchedPath)
+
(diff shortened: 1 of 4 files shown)


python/lsst/daf/persistence/policy.py
```diff
@@ -0,0 +1,252 @@
...
+    initialized with a dict. The only caviat is that keys may NOT contain dots ('.'). This is explained next:
+    Policy extends the dict api so that hierarchical values may be accessed with dot-delimited notiation.
+    That is, foo.getValue('a.b.c') is the same as foo['a']['b']['c'] is the same as foo['a.b.c'], and either
+    of these syntaxes may be used.
+
+    Storage formats supported:
+    - yaml: read and write is supported.
+    - pex policy: read is supported, although this is deprecated and will at some point be removed.
+    """
+
```
ktlim: It's usual to have an `__init__()` method, even if all it does is set `self.data` to something sensible.  You might also want to be able to create a `Policy` directly from a `yaml` or `pex_policy` file.

python/lsst/daf/persistence/policy.py
```diff
@@ -0,0 +1,252 @@
...
+            for key in d.keys():
+                val = d[key]
+                levelKey = base + '.' + key if base is not None else key
+                keys.append(levelKey)
+                if isinstance(val, collections.Mapping):
+                    getKeys(val, keys, levelKey)
+        keys = []
+        getKeys(self.data, keys, None)
+        return keys
+
```
ktlim: I'd like everything from here down to disappear eventually, if not now.  The idea is that we should use native Python interfaces rather than our own C++-like wrappers.

python/lsst/daf/persistence/policy.py
```diff
@@ -0,0 +1,252 @@
...
+            val = [val]
+        elif not isinstance(val, collections.Container):
+            val = [val]
+        return val
+
+
+    ############
+    # yaml i/o #
+    ############
+
```
ktlim: If the inputs were updating an existing `Policy` I'd say they're OK, but as is they might better go into the constructor.  I'd say we should only support one dump format, so the `AsYaml` part of the name is redundant.  If I understand properly, pickling these should be trivial, correct?

python/lsst/daf/persistence/policy.py
n8pease: I could add one, but it would just call the base class __init__. is that desirable?
There are methods to create Policy from yaml, and pex_policy, but it's explicit. I was thinking about creating a version of __init__ that accepts a var and then implements overloading behavior (i.e. dispatch) based on var type. does that seem desirable?

python/lsst/daf/persistence/policy.py
n8pease: makes sense, mostly. there are some intentional differences. big example: .names will return ALL the keys, dot delimited, whereas the normal .keys will return only the 1st level children. So some of the API - that which extends the behavior in a meaningful way - should stay, IMO.

python/lsst/daf/persistence/policy.py
n8pease: ok, that clarifies some of my questions (re. the ctor) above. will do.
pickling - yes.

python/lsst/daf/persistence/policy.py
n8pease: re. explicit methods, now I've seen your comment below. I'll work the ctor.

python/lsst/daf/persistence/policy.py
ktlim: Yes, `names()` should stay (thus the "here down"), but the rest seems not to offer much utility.

python/lsst/daf/persistence/policy.py
n8pease: so there's a polymorphic init now. I suspect the pythonistas will ding me on this as type checking tends to be discouraged in python (as opposed to duck typing) but I think in this case it's justifiable. I'm happy to discuss it though.

python/lsst/daf/persistence/policy.py
```diff
@@ -0,0 +1,392 @@
...
+    # Deprecated methods that mimic pex_policy api.
+    # These are supported (for now), but callers should use the dict api.
+
+    def getValue(self, name):
+        """Get the value for a parameter name/key. See class notes about dot-delimited access.
+
+        :param name:
+        :return: the value for the given name.
+        """
+        warnings.warn_explicit("Deprecated. Use []", DeprecationWarning)
```
n8pease: if there's a better way to mark methods as deprecated I'd like to know about it. These don't actually do anything; the DeprecationWarning is ignored by default. But it's more than nothing at least.

python/lsst/daf/persistence/policy.py
```diff
@@ -0,0 +1,392 @@
...
+import UserDict
+import warnings
+import yaml
+
+import lsst.pex.policy as pexPolicy
+
+class Policy(UserDict.UserDict):
+    """Policy implements a datatype that is used by Butler for configuration parameters.
+    It is essentially a dict with key/value pairs, including nested dicts (as values). In fact, it can be
+    initialized with a dict. The only caviat is that keys may NOT contain dots ('.'). This is explained next:
```
ktlim: "caveat"

python/lsst/daf/persistence/policy.py
```diff
@@ -0,0 +1,392 @@
...
+        - an instance of a Policy
+        - an instance of pex_policy
+        - default policy data with a populated instance of a defaultPolicyInitData
+            Policy.defaultPolicyInitData('daf_persistence', 'testPolicy.yaml', 'tests')
+
+        :param other: data to be loaded
+        :return:
+        """
+        UserDict.UserDict.__init__(self)
+        self.__init(other)
```
ktlim: Is this really necessary?  Why not just merge the next function into this one?

python/lsst/daf/persistence/policy.py
```diff
@@ -0,0 +1,392 @@
...
+
+    defaultPolicyInitData = collections.namedtuple('defaultPolicyInitData', 'productName filePath repos')
+    # Contains data for loading a default policy from a package:
+    # product name: the name of the pacage
+    # filePath: the name of the policy file
+    # the file path from the root of the directory to the dir that contains the file.
+
+    def __init__(self, other=None):
+        """Constructor. Accepts an optional initializer var which may be one of:
+        - a path to a yaml or paf configuration file
```
ktlim: You could possibly make these separate keyword args with defaults of None and avoid a typeswitch (but then you need to define a precedence order in case more than one is specified).

python/lsst/daf/persistence/policy.py
```diff
@@ -0,0 +1,392 @@
...
+        :param extensionPreference: the order in which to try to open files. Will use the first one that
+        succeeds.
+        :return:
+        """
+        extensionPreference = ('yaml', 'paf')
+        matchedPath = None
+        if isinstance(paths, basestring):
+            paths = [paths]
+        for ext in extensionPreference:
+            if matchedPath is not None:
```
ktlim: Is this a leftover?  How can it be anything but None?

python/lsst/daf/persistence/policy.py
```diff
@@ -0,0 +1,392 @@
...
+                break
+            for path in paths:
+                if path.endswith(ext):
+                    matchedPath = path
+                    break
+        policy = None
+        if matchedPath is None:
+            policy = Policy() # return a default policy? throw? WTD?
```
ktlim: If you were given an invalid argument (a policy file path without a valid extension), throwing would be appropriate.

python/lsst/daf/persistence/policy.py
```diff
@@ -0,0 +1,392 @@
...
+    def __initFromDefaultPolicy(self, productName, filePath, repos):
+        """Open a policy file by package name + location. Usually used for package-default policies.
+
+        :param productName: the name of the product that contains the policy.
+        :param filePath: the relative pathname (including file name) to the policy file.
+        :param repos: optional. The subdirectory within the product where policy files are stored. If empty,
+        the product install dir + filePath will be used.
+        :return:
+        """
+        policyPath = os.path.join(os.environ[productName.upper() + '_DIR'], repos, filePath)
```
ktlim: Better to use lsst.utils.getPackageDir() here.

python/lsst/daf/persistence/policy.py
```diff
@@ -0,0 +1,392 @@
...
+                if path.endswith(ext):
+                    matchedPath = path
+                    break
+        policy = None
+        if matchedPath is None:
+            policy = Policy() # return a default policy? throw? WTD?
+        elif matchedPath.endswith('yaml'):
+            self.__initFromYamlFile(matchedPath)
+        elif matchedPath.endswith('paf'):
+            policy = pexPolicy.Policy_createPolicy(matchedPath)
```
ktlim: You should never have to use {class}_{name} SWIG-wrapped methods.  You should use `pexPolicy.Policy.createPolicy()` here, I think.

python/lsst/daf/persistence/policy.py
```diff
@@ -0,0 +1,392 @@
...
+                        d[k] = r
+                    else:
+                        d[k] = u[k]
+                else:
+                    d = {k: u[k]}
+            return d
+        doUpdate(self.data, other)
+
+    def merge(self, other):
+        """Like Policy.update, but will add keys & vales from other that DO NOT EXIST in self. Keys and
```
ktlim: "values"

python/lsst/daf/persistence/policy.py
```diff
@@ -0,0 +1,392 @@
...
+        values that already exist in self will NOT be overwritten.
+
+        :param other:
```
ktlim: Should note that `other` is modified!

python/lsst/daf/persistence/policy.py
```diff
@@ -0,0 +1,392 @@
...
+        :return:
+        """
+        other.update(self)
+        self.data = other.data
+
+    def getSubPolicyVal(self, subPolicyName, subPolicyKey):
+        """Look for a subpolicy and if found, return the value for a key.
```
ktlim: Where is this used?  I don't recall needing something like this, and it seems a bit dangerous as you don't know (and can't tell afterwards) where in the hierarchy it will be found.

python/lsst/daf/persistence/policy.py
n8pease: it's used in __initFromDefaultPolicy. maybe there's another way, but this worked.

python/lsst/daf/persistence/policy.py
n8pease: that's not swigged. I suppose I should be using a PolicySource. How to do that is not immediately obvious. If it seems really important I can work on it. If it's much work I wonder if it's worth "doing it right" if we're going to deprecate and then remove pex policy very soon.

python/lsst/daf/persistence/policy.py
n8pease: good point. probably should not be modifying other anyway. I'll make a copy.

python/lsst/daf/persistence/policy.py
n8pease: I didn't like it either. There was a specific use of it, which I can't find now. I'll take it out, and if jenkins fails again I'll have more info.

PR #10, opened 2016-03-04
butler improvements and concept for finding repositories via butler

n8pease:
and concept for finding repositories via butler


n8pease:
moving this PR from tickets/DM-4168 to tickets/DM-5452. the latter represents foundational work, the former represents use-case-specific implementation

PR #13, opened 2016-03-23
repository returns single element instead of single-item list

n8pease:


PaulPrice:
Commit should be in the imperative, rather than describing what it fixes.

python/lsst/daf/persistence/repository.py
```diff
@@ -333,6 +333,8 @@ def doParents(self, func, *args, **kwargs):
 
         if len(ret) is 0:
             ret = None
+        elif len(ret) is 1:
```
ktlim: While this is reasonable, I think I'd also like something around (post-PR) line 361 that merges the dictionaries if there is a list returned from this method (because there are multiple parents).

PR #14, opened 2016-03-25
fix use of comparison opeators

n8pease:
changed 'is' and 'is not' to == and != where appropriate

### DIFF
--- a/python/lsst/daf/persistence/repository.py
+++ b/python/lsst/daf/persistence/repository.py
@@ -292,7 +292,7 @@ class Repository(object):
                     ret.extend(res)
                 except TypeError:
                     ret.append(res)
-        if len(ret) is 0:
+        if len(ret) == 0:
             ret = None
         return ret
 
--- a/python/lsst/daf/persistence/repositoryMapper.py
+++ b/python/lsst/daf/persistence/repositoryMapper.py
@@ -66,7 +66,7 @@ class RepositoryMapper(Mapper):
         # todo check: do we need keys to complete dataId? (search Registry)
         template = self.policy['repositories.cfg.template']
         location = template % dataId
-        if write is False and self.access.storage.exists(location) is False:
+        if write == False and self.access.storage.exists(location) == False:
             return None
         bl = ButlerLocation(
             pythonType = self.policy['repositories.cfg.python'],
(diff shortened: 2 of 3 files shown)


python/lsst/daf/persistence/repository.py
```diff
@@ -292,7 +292,7 @@ def doSelfAndPeers(self, func, *args, **kwargs):
                     ret.extend(res)
                 except TypeError:
                     ret.append(res)
-        if len(ret) is 0:
+        if len(ret) == 0:
```
brianv0: In general I think Idiomatic python is:

```python
if not len(some_list): # no length
    pass

if len(some_list): # length
    pass

if some_str == "something": 
    pass

if some_str: # Implies len(some_str) > 0
    pass

if some_num == 42:
    pass

if maybe_true:
    pass

if not maybe_true:
    pass


if maybe_none_or_false:
    pass
elif maybe_none_or_false is None:
    pass
else:    # implies maybe_non_or_false == False
   pass


if maybe_none_or_false:
    pass
elif maybe_none_or_false == False:
    pass
else:    # implies maybe_non_or_false is None
   pass

```

python/lsst/daf/persistence/repositoryMapper.py
```diff
@@ -66,7 +66,7 @@ def map_cfg(self, dataId, write):
         # todo check: do we need keys to complete dataId? (search Registry)
         template = self.policy['repositories.cfg.template']
         location = template % dataId
-        if write is False and self.access.storage.exists(location) is False:
+        if write == False and self.access.storage.exists(location) == False:
```
brianv0: `if not write and not self.access.storage.exists(location)`

python/lsst/daf/persistence/repositoryMapper.py
```diff
@@ -79,7 +79,7 @@ def map_cfg(self, dataId, write):
 
 
     def map_repo(self, dataId, write):
-        if write is True:
+        if write == True:
```
brianv0: This should definitely be just `if write:`

timj:
I think this PR should be closed as the changes have been merged in 20fccfa82560b10a3c0fe675a01d1b7cbe4a829a

PR #16, opened 2016-05-24
Replace boost::scoped_array with std::unique_ptr

jdswinbank:


src/DbStorageImpl.cc
```diff
@@ -50,6 +50,7 @@ static char const* SVNid __attribute__((unused)) = "$Id$";
 #include <stdlib.h>
 #include <unistd.h>
 #include <vector>
+#include <memory>
```
jdswinbank: Should be alphabetically sorted ([coding standard 4-10](http://developer.lsst.io/en/latest/coding/cpp_style_guide.html#include-statements-should-be-sorted-and-grouped)).

src/DbAuth.cc
```diff
@@ -43,9 +43,9 @@ static char const* SVNid __attribute__((unused)) = "$Id$";
 #include "lsst/daf/base/Citizen.h"
 #include "lsst/daf/persistence/DbAuth.h"
 
-#include "boost/scoped_array.hpp"
 #include <cstdlib>
 #include <fstream>
+#include <memory>
```
jdswinbank: Maybe worth updating the copyright date in this file (and elsewhere) when you touch it.

PR #17, opened 2016-06-10
Tickets/dm 6459

n8pease:


python/lsst/daf/persistence/butler.py
```diff
@@ -136,44 +149,106 @@ def __init__(self, root, mapper=None, **mapperArgs):
...
+            outputs = [outputs]
+        
+        self.inputs = []
+        self.outputs = [] 
+
+        for cfg in outputs:
+            if cfg['mode'] == 'r':
+                raise RuntimeError('Output repositories must not be read only.')
+            # All inputs become parents of each output, so add them to each output cfg:
+            cfg['parentCfgs'].extend(inputs)
```
ktlim: Is there any problem if the output repo already has the input repo as a parent?

python/lsst/daf/persistence/butler.py
```diff
@@ -136,44 +149,106 @@ def __init__(self, root, mapper=None, **mapperArgs):
...
+            # Outputs must not be read-only:
+            if cfg['mode'] == 'r':
```
ktlim: This looks redundant with lines 196-197.

python/lsst/daf/persistence/butler.py
```diff
@@ -136,44 +149,106 @@ def __init__(self, root, mapper=None, **mapperArgs):
...
+            self._addInputs(cfg)
+
+        for repoData in self.outputs + self.inputs:
+            repoData.repo = Repository.makeFromCfg(repoData.cfg)
+
+    def _addInputs(self, cfg, tags=[]):
+        """Add a RepoData for cfg if it is not already represented in self.inputs."""
+        # Inputs must not be write-only:
+        if cfg['mode'] == 'w':
+            raise RuntimeError('Input repositoires must not be write only.')
```
ktlim: Typo: repositoires -> repositories

python/lsst/daf/persistence/butler.py
```diff
@@ -136,44 +149,106 @@ def __init__(self, root, mapper=None, **mapperArgs):
...
+        # If there is already an output RepoData instance for a given cfg, don't create a new one. Add it to
+        # the inputs list if it's not there, and update its tags.
+        for d in self.outputs:
+            if d.cfg == cfg:
+                self.inputs.append(d)
+                repoData = d
+
+        # If there is already an input RepoData instance for a given cfg, don't create a new one, just update 
+        # its tags.
+        if not repoData:
```
ktlim: Better is `if repoData is None:`.

python/lsst/daf/persistence/butler.py
```diff
@@ -136,44 +149,106 @@ def __init__(self, root, mapper=None, **mapperArgs):
...
+                    repoData = d
+                    break
+
+        # If no repoData exists for the cfg yet, then create a new one.
+        if not repoData:
+            repoData = RepoData(cfg, tags)
+            self.inputs.append(repoData)
+        else:
+            repoData.tags.update(tags)
+        # add the parents to the input list
```
ktlim: You might document here that this effectively sets the search order for inputs to be depth-first.  There's also a lot of wasted effort because all of the original input repos are added as parents to each output repo before the readable output repos are passed to this method: each original input repo (and all of its parents) has one addition attempt for each readable output repo plus one for itself as an input repo, only one attempt of which will actually do anything (except possibly for adding tags).

python/lsst/daf/persistence/butler.py
```diff
@@ -229,35 +304,49 @@ def getKeys(self, datasetType=None, level=None):
...
 
         if format is None:
             format = (key,)
         elif not hasattr(format, '__iter__'):
             format = (format,)
 
-        tuples = self.repository.queryMetadata(datasetType, format, dataId)
+        tuples = None
+        for repoData in self.inputs:
+            if not dataId.tag or len(dataId.tag.intersection(repoData.tags)) > 0:
```
ktlim: I _think_ it should be an error if the same dataset type has different dataId keys in differently-tagged repos, but perhaps it's best to leave this flexibility in for possible future use.

python/lsst/daf/persistence/butler.py
```diff
@@ -229,35 +304,49 @@ def getKeys(self, datasetType=None, level=None):
         @returns (dict) valid data id keys; values are corresponding types.
         """
         datasetType = self._resolveDatasetTypeAlias(datasetType)
-        return self.repository.getKeys(datasetType, level)
+
+        keys = None
+        for repoData in self.inputs:
+            if not tag or len(tag.intersection(repoData.tags)) > 0:
```
ktlim: I think it should be an error if the same dataset type has different dataId keys in differently-tagged repos, but perhaps it's best to leave this flexibility in for possible future use.

python/lsst/daf/persistence/butler.py
```diff
@@ -275,26 +364,25 @@ def queryMetadata(self, datasetType, format=None, dataId={}, **rest):
...
-        try:
-            if len(locations) != 1:
-                raise RuntimeError("Multiple (or none) locations for datasetExists(%s, %s)" %(datasetType, dataId))
-            location = locations[0]
-        except TypeError:
-            # locations might not be a list; that's ok.
-            location = locations
 
         additionalData = location.getAdditionalData()
         storageName = location.getStorageName()
```
ktlim: Eventually I would hope that this PosixStorage-specific determination of existence should move into that class.

python/lsst/daf/persistence/butler.py
```diff
@@ -328,22 +416,19 @@ def get(self, datasetType, dataId={}, immediate=False, **rest):
         """
 
         datasetType = self._resolveDatasetTypeAlias(datasetType)
-        dataId = copy.copy(dataId)
+        dataId = DataId(dataId)
         dataId.update(**rest)
 
-        locations = self.repository.map(datasetType, dataId)
-        if locations is None:
+        location = None
```
ktlim: Redundant.

python/lsst/daf/persistence/butler.py
```diff
@@ -440,13 +525,11 @@ def dataRef(self, datasetType, level=None, dataId={}, **rest):
...
         datasetType = self._resolveDatasetTypeAlias(datasetType)
+        dataId = DataId(dataId)
         subset = self.subset(datasetType, level, dataId, **rest)
         if len(subset) != 1:
-            raise RuntimeError, """No unique dataset for:
-    Dataset type = %s
-    Level = %s
-    Data ID = %s
-    Keywords = %s""" % (str(datasetType), str(level), str(dataId), str(rest))
+            raise RuntimeError("No unique dataset for: Dataset type:%s Level:%s Data ID:%s Keywords:%s" %
```
ktlim: The multi-line exception may be easier to read than the single-line form, but I'm OK with either.

python/lsst/daf/persistence/butler.py
```diff
@@ -479,7 +563,28 @@ def _resolveDatasetTypeAlias(self, datasetType):
...
+        If x is a string, will treat the string as a single object (i.e. not as a list of chars)"""
+        if x is None:
+            x = set()
+        if isinstance(x, basestring):
+            x = set([x])
+        else:
+            try:
+                x = set(x)
+            except TypeError:
+                x= set([x])
```
ktlim: Spacing.

python/lsst/daf/persistence/butlerFactory.py
```diff
@@ -26,7 +26,7 @@
```
ktlim: Do we still need this class at all?  I think all the places that use it can be converted to just the `Butler()` constructor.

python/lsst/daf/persistence/butlerSubset.py
```diff
@@ -238,7 +242,10 @@ def subItems(self, level=None):
         """
 
         if level is None:
-            mappers = self.butlerSubset.butler.repository.mappers()
+            mappers = []
+            for repoData in self.butlerSubset.butler.inputs + self.butlerSubset.butler.outputs:
+                if repoData.repo._mapper not in mappers:
+                    mappers.append(repoData.repo._mapper)
             if len(mappers) != 1:
                 raise RuntimeError("Support for multiple repositories not yet implemented!")
```
ktlim: At least this should say "Support for multiple mappers", since this PR is about multiple repositories :smiley:.  It might be possible to use multiple mappers if they all had the same default level.  (I'm a little concerned about the "todo" comment below, as well.)

python/lsst/daf/persistence/dataId.py
```diff
@@ -0,0 +1,62 @@
...
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+
+import copy
+import UserDict
+
+class DataId(UserDict.IterableUserDict):
+    """DataId is used to pass scientifically meaninful key-value pairs. It may be tagged as applicable only
```
ktlim: Typo: "meaniful" -> "meaningful"

python/lsst/daf/persistence/dataId.py
```diff
@@ -0,0 +1,62 @@
...
+    repositories that are tagged with the same value"""
+
+    def __init__(self, initialdata=None, tag=None, **kwargs):
+        """Constructor
+
+        Parameters
+        -----------
+        initialdata : dict
+            A dict of inital data for the DataId
```
ktlim: Typo: "inital" -> "initial"

python/lsst/daf/persistence/dataId.py
```diff
@@ -0,0 +1,62 @@
...
+class DataId(UserDict.IterableUserDict):
+    """DataId is used to pass scientifically meaninful key-value pairs. It may be tagged as applicable only
+    repositories that are tagged with the same value"""
+
+    def __init__(self, initialdata=None, tag=None, **kwargs):
+        """Constructor
+
+        Parameters
+        -----------
+        initialdata : dict
```
ktlim: initialdata can also be a `DataId`.

python/lsst/daf/persistence/dataId.py
```diff
@@ -0,0 +1,62 @@
...
+        UserDict.UserDict.__init__(self, initialdata)
+        try:
+            self.tag = copy.deepcopy(initialdata.tag)
+        except AttributeError:
+            self.tag = set()
+
+        if tag is not None:
+            if isinstance(tag, basestring):
+                self.tag.update([tag])
+            try:
```
ktlim: This needs to be in an `else` clause, I believe, which also means that test coverage is inadequate here.

python/lsst/daf/persistence/policy.py
```diff
@@ -376,6 +376,36 @@ def getStringArray(self, key):
             val = [val]
         return val
 
+    def __lt__(self, other):
```
ktlim: Does it really make sense to be doing these types of comparisons on Policies?

python/lsst/daf/persistence/posixStorage.py
```diff
@@ -102,20 +85,23 @@ def getMapperClass(root):
             if os.path.exists(os.path.join(basePath, "_parent")):
```
ktlim: Note that `_parent` needs to be supported until repository configs are routinely persisted and the persisted configs are routinely used to construct repositories.

python/lsst/daf/persistence/repository.py
```diff
@@ -200,18 +183,37 @@ def _initMapper(self, repoCfg):
...
+                raise RuntimeError("Mapper is unspecified and parents have different mappers.")
+            elif len(parentMappers) == 0:
                 return None
+            return parentMappers.pop()
+
+        if mapper is None:
+            mapper = inferMapper(repoCfg)
+
+        if mapper is None:
+            raise RuntimeError("Can not determine a mapper to use.")
```
ktlim: Typo: "Can not" -> "Cannot"

python/lsst/daf/persistence/repository.py
```diff
@@ -233,15 +235,20 @@ def _initMapper(self, repoCfg):
                 mapperArgs = copy.copy(repoCfg['mapperArgs'])
                 if mapperArgs is None:
                     mapperArgs = {}
+                # so that root doesn't have to be redundantly passed in cfgs, if root is specified in the
+                # storage and if it is an argument to the mapper, make sure that it's present in mapperArgs.
                 if 'root' not in mapperArgs:
-                    mapperArgs['root'] = self._access.root()
+                    for c in inspect.getmro(mapper):
```
ktlim: This is really pretty ugly.  I think you would be much better off defining some class attributes or methods on `Mapper` that derived classes could use to indicate that they need certain arguments.  Yes, this means changing existing `Mapper` subclasses, but I think all of them are in tests or are `CameraMapper` subclasses.

python/lsst/daf/persistence/repository.py
```diff
@@ -356,44 +291,16 @@ def getKeys(self, *args, **kwargs):
...
-
-        :param args:
-        :param kwargs:
         :return: A dict of {key:valueType}
         """
         # todo: getKeys is not in the mapper API
         if self._mapper is None:
             return None
-        return self._mapper.getKeys(*args, **kwargs)
+        keys = self._mapper.getKeys(*args, **kwargs)
```
ktlim: Why is this better than the existing single line?  Were you thinking of doing something to the keys?

tests/repository.py
```diff
@@ -383,71 +364,250 @@ def tearDown(self):
...
+                                     storageCfg=dp.PosixStorage.cfg(root='tests/repository/repoB'), 
+                                     mapper=TestMapperInference.AlternateMapper)
+        repoCCfg = dp.Repository.cfg(mode='w', 
+                                     storageCfg=dp.PosixStorage.cfg(root='tests/repository/repoC'), 
+                                     parentCfgs=(repoACfg, repoBCfg))
+        self.assertRaises(RuntimeError, dp.Butler, outputs=repoCCfg)
+
+
 
 def suite():
```
ktlim: See https://developer.lsst.io/coding/python_testing.html for the new way to do this.

python/lsst/daf/persistence/butler.py
n8pease: I don't mean is not None though: e.g. "set() is None" is False, but "not set()" is True

python/lsst/daf/persistence/butler.py
n8pease: I was aware it was inefficient, but it's simple and I was preferring that. I don't think this particular operation is going to be performed frequently e.g. in an inner loop so I wasn't worried about some efficiency loss. Do you think it's worth optimizing?

python/lsst/daf/persistence/policy.py
n8pease: it had to do with putting these into a container, these rich comparison methods were needed for that.

python/lsst/daf/persistence/posixStorage.py
n8pease: I don't expect that this _parent code will go away any time soon. I think I need to add a test and impl code here for the new style cfgs (when a repo has a sparsely populated cfg and a parent cfg that defines its mapper)

I don't know about line that says `globals = {}`, I'm fairly certain it was copied from butler when I created the posixStorage class. I'll look into it though.

python/lsst/daf/persistence/repository.py
n8pease: http://www.oxforddictionaries.com/words/cannot-or-can-not but ok

python/lsst/daf/persistence/repository.py
n8pease: assigning keys to a variable before returning it improves the ability to debug it. I needed it for something.

python/lsst/daf/persistence/butler.py
n8pease: I expect it to be part of DM-6225

python/lsst/daf/persistence/posixStorage.py
n8pease: `globals = {}` did come over from an old version of Butler. I can't see any reason why paving over the globals keyword in this context is good and tests seem to run fine without it so I'll take it out.

python/lsst/daf/persistence/butlerFactory.py
n8pease: I can do that but it'll touch maybe a dozen packages, and I think it would be more manageable to do it as part of a separate ticket.

python/lsst/daf/persistence/butler.py
```diff
@@ -61,6 +55,66 @@ class ButlerCfg(Policy, yaml.YAMLObject):
...
+            return self._inputs
+        self._inputs = [rd for rd in self.byRepoRoot.itervalues() if 'r' in rd.mode]
+        return self._inputs
+    def outputs(self):
+        if self._outputs is not None:
+            return self._outputs
+        self._outputs = [rd for rd in self.byRepoRoot.itervalues() if 'w' in rd.mode]
+        return self._outputs
+    def all(self):
+        self._all = [rd for rd in self.byRepoRoot.itervalues()]
```
ktlim: Add `if self._all is not None:` here?

python/lsst/daf/persistence/butler.py
```diff
@@ -136,44 +190,165 @@ def __init__(self, root, mapper=None, **mapperArgs):
...
+        for args in outputs:
+            self._addRepo(args, inout='out', defaultMapper=defaultMapper)
+
+        for args in inputs:
+            self._addRepo(args, inout='in')
+
+        for repoData in self._repos.outputs():
+            inputCfgRoots = [rd.args.cfgRoot for rd in self._repos.inputs() if rd is not repoData]
+            repoData.cfg.addParents(inputCfgRoots)
+            Storage.putRepositoryCfg(repoData.cfg, repoData.args.cfgRoot)
```
ktlim: This overwrites the existing configuration with the new parents, correct?  A comment to that effect might be good.  I'm also worried about how all of these `put`s are going to work when run in parallel across many nodes.  Please make sure that things like `matchesArgs()` will work properly if some nodes see the parents added and some do not.

python/lsst/daf/persistence/repository.py
```diff
@@ -29,231 +29,100 @@
...
+            # so that root doesn't have to be redundantly passed in cfgs, if root is specified in the
+            # storage and if it is an argument to the mapper, make sure that it's present in mapperArgs.
+            for arg in ('root', 'storage'):
+                if arg not in mapperArgs:
+                    mro = inspect.getmro(mapper)
+                    if mro[-1] is object:
+                        mro = mro[:-1]
+                    for c in mro:
+                        if arg in inspect.getargspec(c.__init__ ).args:
+                            mapperArgs[arg] = self._storage.root
```
ktlim: The Mapper gets the root as the value of `storage`?  This seems a little odd.

python/lsst/daf/persistence/butler.py
```diff
@@ -39,15 +39,9 @@
 import lsst.pex.logging as pexLog
 import lsst.pex.policy as pexPolicy
 from lsst.daf.persistence import StorageList, LogicalLocation, ReadProxy, ButlerSubset, ButlerDataRef, \
```
ktlim: This almost seems like you could `import *`, since it's the same module.  Otherwise it's a long list...

python/lsst/daf/persistence/safeFileIo.py
```diff
@@ -48,6 +50,40 @@ def setFileMode(filename):
...
+    with tempfile.NamedTemporaryFile(dir=outDir, prefix=outName, delete=False) as temp:
+        try: 
+            yield temp
+        finally:
+            try:
+                temp.close()
+                os.symlink(temp.name, name)
+                # If the symlink was created then this is the process that created the first instance of the 
+                # file, and we know its contents match. Move the temp file over the symlink.
+                if os.path.islink(name):
```
ktlim: If this is False, something bad happened, no?

python/lsst/daf/persistence/safeFileIo.py
```diff
@@ -48,6 +50,40 @@ def setFileMode(filename):
...
+    is silently thrown away. If they are not the same then a runtime error is raised.
+    """
+    outDir, outName = os.path.split(name)
+    safeMakeDir(outDir)
+    with tempfile.NamedTemporaryFile(dir=outDir, prefix=outName, delete=False) as temp:
+        try: 
+            yield temp
+        finally:
+            try:
+                temp.close()
```
ktlim: Could this plus the `with` statement lead to two `close()` calls?  Is there any chance that could be unsafe?

python/lsst/daf/persistence/safeFileIo.py
```diff
@@ -48,6 +50,40 @@ def setFileMode(filename):
...
+
+    The context manager provides a temporary file object. After the user is done, the temporary file becomes
+    the permanant file if the file at name does not already exist. If the file at name does exist the 
+    temporary file is compared to the file at name. If they are the same then this is good and the temp file
+    is silently thrown away. If they are not the same then a runtime error is raised.
+    """
+    outDir, outName = os.path.split(name)
+    safeMakeDir(outDir)
+    with tempfile.NamedTemporaryFile(dir=outDir, prefix=outName, delete=False) as temp:
+        try:
```
ktlim: Why the `try:` block here?

python/lsst/daf/persistence/safeFileIo.py
```diff
@@ -1,4 +1,4 @@
-#
+    #
```
ktlim: Errant whitespace.

python/lsst/daf/persistence/safeFileIo.py
```diff
@@ -48,6 +50,40 @@ def setFileMode(filename):
...
+                temp.close()
+                os.symlink(temp.name, name)
+                # If the symlink was created then this is the process that created the first instance of the 
+                # file, and we know its contents match. Move the temp file over the symlink.
+                if os.path.islink(name):
+                    os.rename(temp.name, name)
+            except OSError as e:
+                if e.errno != errno.EEXIST:
+                    raise e
+                filesMatch = filecmp.cmp(temp.name, name, shallow=False)
```
ktlim: There is the possibility that this could get more complicated.  For example, there could be randomization of the order of Config items or headers or something.  We might need to pass in a comparison function in some cases.

python/lsst/daf/persistence/safeFileIo.py
n8pease: try/finally is for the yield. 
try/except OSError is for creating (or not) the symlink.

python/lsst/daf/persistence/safeFileIo.py
n8pease: good catch. calling close() on a file object twice does not seem to be a problem. However, it's redundant. I want to close the file explicitly before making the symlink and moving the file so I took out the ``with`` statement.

python/lsst/daf/persistence/repository.py
n8pease: eventually the mapper should have no concept of root. we will be working towards that over the next year.

python/lsst/daf/persistence/__init__.py
```diff
@@ -22,16 +22,20 @@
 
 """Python interface to lsst::daf::persistence classes
 """
+from utils import *
```
TallJimbo: I think these should all be relative (`from .utils import *`) imports, if they refer to the same package.  That may not matter now, but I think it's the convention we use in most other packages and it may be important for Python 3.

python/lsst/daf/persistence/butler.py
TallJimbo: I'm not actually sure it's safe to import directly from the package a module lives in; it best it seems fragile against moving that module up or down in the list of modules imported in the package's `__init__` file (and I'm actually a bit surprised it works at all).

python/lsst/daf/persistence/butler.py
```diff
@@ -61,6 +55,65 @@ class ButlerCfg(Policy, yaml.YAMLObject):
     def __init__(self, cls, repoCfg):
         super(ButlerCfg, self).__init__({'repoCfg':repoCfg, 'cls':cls})
 
+class RepoData(object):
```
TallJimbo: Docstring?

python/lsst/daf/persistence/butler.py
```diff
@@ -61,6 +55,65 @@ class ButlerCfg(Policy, yaml.YAMLObject):
...
+        self.repo = repo
+        self.mode = args.mode
+        # self.tags is used to keep track of *all* the applicable tags to the Repo, not just the tags in
+        # the cfg (e.g. parents inherit their childrens' tags)
+        if not isinstance(tags, set):
+            raise RuntimeError("tags passed into RepoData must be in a set.")
+        self.tags = tags
+
+    def __reduce__(self):
+        return (_unreduceRepoData, (self.args, self.cfg, self.repo, self.mode, self.tags))
```
TallJimbo: I think you can just use the `RepoData` type itself as the callable here instead of adding a new function to unreduce.

python/lsst/daf/persistence/butler.py
```diff
@@ -61,6 +55,65 @@ class ButlerCfg(Policy, yaml.YAMLObject):
...
+    def __repr__(self):
+        return "RepoData(args=%s cfg=%s repo=%s tags=%s" % (self.args, self.cfg, self.repo, self.tags)
+
+
+def _unreduceRepoData(args, cfg, repo, tags):
+    repoData = RepoData(RepositoryArgs(args, cfg, repo, tags))
+    return repoData
+
+
+class RepoDataContainer(object):
```
TallJimbo: Class and methods need docstrings.  I think conventions say there should be newlines between methods.

python/lsst/daf/persistence/butler.py
```diff
@@ -136,44 +175,213 @@ def __init__(self, root, mapper=None, **mapperArgs):
...
+        @param defaultMapper (mapper class or None ) if a default mapper could be inferred from inputs then 
+                                                     this will be a class object that can be used for any 
+                                                     outputs that do not explicitly define their mapper. If 
+                                                     None then a mapper class could not be inferred and a 
+                                                     mapper must be defined by each output.
+        @param butlerIOParents (ordered dict) The keys are cfgRoot of repoArgs, val is the repoArgs. 
+                                              This is all the explicit input and output repositories to the
+                                              butler __init__ function, it is used when determining what the
+                                              parents of writable repositories are.
+        (list of string) URI to parent cfg, of repositories that are directly passed
```
TallJimbo: It looks like this documentation line may not correspond to any argument.

python/lsst/daf/persistence/butler.py
```diff
@@ -136,44 +175,213 @@ def __init__(self, root, mapper=None, **mapperArgs):
...
+                    # repositories used by this butler. (If there is a configuration change a new output 
+                    # repository should be created). 
+                    # IE cfg.parents can have parents that are not passed in as butler inputs or readable 
+                    # outputs, but the butler may not have readable i/o that is not a parent of an already 
+                    # existing output.
+                    for cfgRoot in butlerIOParents:
+                        if cfgRoot not in cfg.parents:
+                            if cfgRoot != args.cfgRoot or 'r' not in args.mode:
+                                raise RuntimeError(
+                                    "Existing output repository parents do not match butler's inputs.")
```
TallJimbo: I don't understand `or 'r' not in args.mode` here.  If we're setting up an output repository, shouldn't this check do the same thing regardless of whether or not it's readable as well as writeable?

python/lsst/daf/persistence/butler.py
```diff
@@ -136,44 +175,213 @@ def __init__(self, root, mapper=None, **mapperArgs):
...
+    @staticmethod
+    def getDefaultMapper(inputs):
+        mappers = set()
+        for args in inputs:
+            if args.mapper is not None:
+                mappers.add(args.mapper)
+            else:
+                cfgRoot = args.cfgRoot
+                mapper = Butler.getMapperClass(cfgRoot)
+                mappers.add(mapper)
```
TallJimbo: If I'm reading this right, `args.mapper` is either unsanitized or an instance at this stage, which I don't think we can expect to be hashable, right?  Or is that a new requirement on mappers?

python/lsst/daf/persistence/butler.py
```diff
@@ -229,35 +437,49 @@ def getKeys(self, datasetType=None, level=None):
         @returns (dict) valid data id keys; values are corresponding types.
```
TallJimbo: Need to add `tag` to docstring.

python/lsst/daf/persistence/butler.py
```diff
@@ -229,35 +437,49 @@ def getKeys(self, datasetType=None, level=None):
...
         """
         datasetType = self._resolveDatasetTypeAlias(datasetType)
-        return self.repository.getKeys(datasetType, level)
+
+        keys = None
+        for repoData in self._repos.inputs():
+            if not tag or len(tag.intersection(repoData.tags)) > 0:
```
TallJimbo: Are we certain `tag` is a set here?  This looks like a public method, and docs seem to suggest tags can be anything.

python/lsst/daf/persistence/dataId.py
```diff
@@ -0,0 +1,63 @@
...
+    def __init__(self, initialdata=None, tag=None, **kwargs):
+        """Constructor
+
+        Parameters
+        -----------
+        initialdata : dict or dataId
+            A dict of initial data for the DataId
+        tag : any type, or a container of any type
+            A value or container of values used to restrict the DataId to one or more repositories that 
+            share that tag value.
```
TallJimbo: Please document (maybe not here) that while you can pass anything in as a tag, `self.tag` is always a `set`.

python/lsst/daf/persistence/butler.py
```diff
@@ -229,35 +437,49 @@ def getKeys(self, datasetType=None, level=None):
...
+        @param **rest keyword arguments for the partial data id.
         @returns (list) a list of valid values or tuples of valid values as
         specified by the format (defaulting to the same as the key) at the
         key's level of granularity.
         """
 
         datasetType = self._resolveDatasetTypeAlias(datasetType)
-        dataId = copy.copy(dataId)
+        dataId = DataId(dataId)
         dataId.update(**rest)
```
TallJimbo: Does this set the tag correctly if it's included in the extra kwargs in `**rest`?

python/lsst/daf/persistence/butlerSubset.py
```diff
@@ -29,6 +29,8 @@
 
 from __future__ import with_statement
 
+from lsst.daf.persistence import DataId
```
TallJimbo: Should probably be a relative import.

python/lsst/daf/persistence/repository.py
```diff
@@ -29,231 +29,107 @@
...
+        return "%s(root=%r, cfgRoot=%r, mapper=%r, mapperArgs=%r, tags=%s, mode=%r)" % (
+            self.__class__.__name__, self.root, self._cfgRoot, self.mapper, self.mapperArgs, self.tags, 
+            self.mode)
+
+    @property
+    def cfgRoot(self):
+        if self._cfgRoot is not None:
+            return self._cfgRoot
+        else:
+            return self.root
```
TallJimbo: This could be rewritten as a conditional statement:
```
return self._cfgRoot if else._cfgRoot is not None else self.root
```

python/lsst/daf/persistence/repository.py
```diff
@@ -29,231 +29,107 @@
...
-    * if parentJoin is 'outer' will return all results found.
-
-    Recursion is implemented in a few functions which may be overridden to alter recursive behavior:
-    def doPeersAndParents()
-    def doPeers()
-    def doParents()
+    def tag(self, tag):
+        """add a tag to the repository cfg"""
+        if isinstance(tag, basestring):
+            self.tags.add(tag)
```
TallJimbo: Should this be `if isinstance(tag, set)`?  I don't think `add` is a method of `basestring`.

python/lsst/daf/persistence/repositoryCfg.py
```diff
@@ -0,0 +1,152 @@
...
+    def v1Constructor(loader, node):
+        """Constructor for 'version 1' of the serlized RepositoryCfg. 
+
+        If new parameters are added to RepositoryCfg they will have to be checked for in d; if they are there
+        then their value should be used and if they are not there a default value must be used in place.
+
+        In case the structure of the serialzed file must be changed in a way that invalidates some of the
+        keys:
+        1. Increment the version number (after _v1) in the yaml_tag of this class.
+        2. Add a new constructor (similar to this one) to deserialze new serilizations of this class.
```
TallJimbo: Typo: `serilizations`

python/lsst/daf/persistence/repositoryCfg.py
```diff
@@ -0,0 +1,152 @@
...
+            self._parents = list(parents)
+        else:
+            self._parents = parents
+        self._isLegacyRepository = isLegacyRepository
+
+    @staticmethod
+    def v1Constructor(loader, node):
+        """Constructor for 'version 1' of the serlized RepositoryCfg. 
+
+        If new parameters are added to RepositoryCfg they will have to be checked for in d; if they are there
```
TallJimbo: What's "d"?

python/lsst/daf/persistence/repositoryCfg.py
```diff
@@ -0,0 +1,152 @@
...
+        return cfg
+
+    def __eq__(self, other):
+        if not other:
+            return False
+        return self._root == other._root and \
+               self.mapper == other._mapper and \
+               self.mapperArgs == other._mapperArgs and \
+               self.parents == other._parents and \
+               self._isLegacyRepository == other._isLegacyRepository
```
TallJimbo: Should define `__ne__` too to avoid surprises.

python/lsst/daf/persistence/repositoryCfg.py
```diff
@@ -0,0 +1,152 @@
...
+    def mapper(self):
+        return self._mapper
+
+    @property
+    def mapperArgs(self):
+        return self._mapperArgs
+    
+    @mapperArgs.setter
+    def mapperArgs(self, newDict):
+        self._mapperArgs = newDict
```
TallJimbo: There doesn't seem to be any reason to make `mapperArgs` a `property`; its getter and setter are trivial.

python/lsst/daf/persistence/repositoryCfg.py
```diff
@@ -0,0 +1,152 @@
...
+                    return False
+            else:
+                if type(self._mapper) != type(repositoryArgs.mapper):
+                    return False
+        # check mapperArgs for any keys in common and if their value does not match then return false.
+        if self._mapperArgs is not None and repositoryArgs.mapperArgs is not None:
+            for key in set(self._mapperArgs.keys()) & set(repositoryArgs.mapperArgs):
+                if self._mapperArgs[key] != repositoryArgs.mapperArgs[key]:
+                    return False
+        return True
```
TallJimbo: Is it possible that `repositoryArgs` and `self` could refer to the same mapper differently (i.e. one via string, one via a class), and hence incorrectly compare as inequal here?

python/lsst/daf/persistence/safeFileIo.py
```diff
@@ -48,6 +50,43 @@ def setFileMode(filename):
     os.chmod(filename, (~umask & 0o666))
 
 @contextmanager
+def FileForWriteOnceCompareSame(name):
+    """Context manager to get a file that can be written only once and all other writes will succeed only if
+    they match the inital write.
+
+    The context manager provides a temporary file object. After the user is done, the temporary file becomes
+    the permanant file if the file at name does not already exist. If the file at name does exist the
```
TallJimbo: typo `permanant` -> `permanent`

python/lsst/daf/persistence/storage.py
```diff
@@ -0,0 +1,124 @@
...
+        .. warning::
+
+            Storage is 'wet paint' and very likely to change during factorization of Butler back end and 
+            storage formats (DM-6225). Use of it in production code other than via the 'old butler' API is 
+            strongly discouraged.
+
+        Parameters
+        ----------
+        scheme : str
+            Name of the `scheme` the class is being registered for, which appears at the beginning of a URI.
```
TallJimbo: Some examples would go a long way towards explaining what a "scheme" is.

python/lsst/daf/persistence/storage.py
```diff
@@ -0,0 +1,124 @@
...
+            storage formats (DM-6225). Use of it in production code other than via the 'old butler' API is 
+            strongly discouraged.
+
+        :param uri: the uri to a locaiton that contains a repositoryCfg.
+        :return: a Storage subclass instance.
+        '''
+        ret = None
+        parseRes = urlparse.urlparse(uri)
+        if parseRes.scheme in Storage.storages:
+            theClass = Storage.storages[parseRes.scheme]
```
TallJimbo: Shouldn't matter in this case, but as a good general practice I'd recommend instead doing
```
theClass = Storage.storages.get(parseRes.scheme, None)
if theClass is None:
    raise ...
```
just so there's only one dict lookup.

python/lsst/daf/persistence/utils.py
```diff
@@ -0,0 +1,58 @@
...
+    """Takes any object and puts that whole object in a list:
+    - strings will be made into a single element in the list
+    - tuples will be converted to list
+    - lists will remain as lists
+    - None will be made into an empty list
+    """
+    if x is None:
+        x = []
+    if not hasattr(x, '__iter__'):
+        x = [x]
```
TallJimbo: Why check for `__iter__` here but use `isinstance(x, basestring)` below?  I think both work, but the latter seems safer.

tests/reposInButler.py
```diff
@@ -34,9 +34,14 @@
...
 import lsst.daf.persistence as dp
-from lsst.daf.persistence import posixRepoCfg, Policy
+from lsst.daf.persistence import Policy
+
+
+def setup_module(module):
+    lsst.utils.tests.init()
+
 
 class PosixPickleStringHanlder:
```
TallJimbo: Not on this changeset, but there's a typo here: `Hanlder`.

tests/repository.py
```diff
@@ -140,69 +130,56 @@ class TestBasics(unittest.TestCase):
...
             shutil.rmtree('tests/repository')
-        del self.butler
+        if os.path.exists('tests/butlerAlias/repositoryCfg.yaml'):
+            os.remove('tests/butlerAlias/repositoryCfg.yaml')
 
     def testGet(self):
-        raw_image = self.butler.get(self.datasetType, {'visit':'2', 'filter':'g'})
+        raw_image = self.butler.get('raw', {'visit':'2', 'filter':'g'})
         # in this case the width is known to be 1026:
         self.assertEqual(raw_image[1].header["NAXIS1"], 1026) # raw_image is an lsst.afw.ExposureU
```
TallJimbo: Not this changeset, but this comment is definitely wrong.  It looks like it may be a PyFITS/astropy.io.fits object.

tests/repository.py
```diff
@@ -383,75 +348,361 @@ def tearDown(self):
...
+        self.assertEqual(butler.get('foo', {'bar':3}), objJkl)
+
+
+class TestTagging(unittest.TestCase):
+    """A test case for the tagging of repository classes.
+    """
+
+    def tearDown(self):
+        if os.path.exists('tests/repository'):
+            shutil.rmtree('tests/repository')
```
TallJimbo: Maybe use a different for this test and the one above it?

TallJimbo: In fact, that may be necessary if we switch to a test runner that can run multiple tests from the same file in parallel.

tests/repository.py
```diff
@@ -383,75 +348,361 @@ def tearDown(self):
...
+
+class MemoryTester(lsst.utils.tests.MemoryTestCase):
+    pass
+
 
 if __name__ == '__main__':
-    run(True)
+    lsst.utils.tests.init()
+    unittest.main()
+
```
TallJimbo: Lots of unnecessary whitespace here.

python/lsst/daf/persistence/butler.py
n8pease: It seems to depend on coupling. IE you can run into problem if A comes before B in the __init__.py file and A imports B. But as long as the classes are coupled properly it works fine.

python/lsst/daf/persistence/butler.py
n8pease: class instances seem to be hashable(?). if no __eq__ method is defined I think it just uses the address, which should be ok here.

python/lsst/daf/persistence/butler.py
n8pease: tag does not have to be set (because of "if not tag").

PaulPrice: Jim said "a `set`", not "is set".

n8pease: ah. so he did. thanks.

python/lsst/daf/persistence/butler.py
PaulPrice:
```
>>> class Foo(object): pass
... 
>>> hash(Foo)
8761989005478
```

python/lsst/daf/persistence/butler.py
n8pease: I'm not sure it would be correct for the tag to be updated by values in ``**rest**``. DataId allows for the ``tag`` key as well as tag var (which could be confusing, but to disallow ``tag`` as a key also seems potentially confusing to me). Do you have a use case for this?

python/lsst/daf/persistence/repositoryCfg.py
n8pease: yes, but I think if it's being specified differently then there is a very good chance something is not being done right in the pipeline. I think we should leave it for now and evaluate as needed.

python/lsst/daf/persistence/storage.py
n8pease: I was choosing brevity over efficiency (I mean it's python...), but you're probably right.

python/lsst/daf/persistence/utils.py
n8pease: it's complicated and has to do with dissimilar behaviors between set and list builders, especially with regard to strings.

python/lsst/daf/persistence/butler.py
TallJimbo: I guess I actually meant the `mappers.add(args.mapper)` up on line 374, not the one on 378 - I agree that classes should be hashable.

But if all of the inputs specify the same mapper class with the same uri, but instantiate those mappers before passing them into the `RepositoryArgs`, those mappers will not have the same address but should compare as equal here.

python/lsst/daf/persistence/butler.py
TallJimbo: The most natural way for users to specify a tag in a butler call (IMO) is something like:
```
calexp = butler.get("calexp", visit=236, ccd=44, tag="cfht")
```
If the intent is to force them to instead do
```
calexp = butler.get("calexp", DataId(visit=236, ccd=44, tag="cfht"))
```
that's not terrible, but it needs to be well-documented to avoid causing confusion.

python/lsst/daf/persistence/repositoryCfg.py
TallJimbo: Another option (both here and in the hashability discussion elsewhere on this PR) is to simply reduce the ways a mapper can be specified.  If the the task framework specifying different mappers in different ways is effectively being considered a bug here, we should just disallow some of those ways up front and force the task framework to be consistent.  Of course, if some of these mapper arg types are just for backwards compatibility, that's different, but I didn't get that impression.

python/lsst/daf/persistence/utils.py
TallJimbo: In that case a code comment would be welcome.

python/lsst/daf/persistence/repositoryCfg.py
n8pease: to remain backward compatible we do need to allow class instance, class object, and importable string (to class object).

PR #19, opened 2016-07-29
DM-7069: Port to Python 3

timj:
This PR is for Python 3 work and other cleanups.


tests/butlerAlias.py
```diff
@@ -65,7 +65,7 @@ def getDefaultLevel(self):
         return 'visit'
 
     def getKeys(self, datasetType, level):
-        return {'filter': types.StringType, 'visit': types.IntType}
+        return {'filter': bytes, 'visit': int}
```
n8pease: why bytes here and not str? (seems wrong, but maybe it's a new python3 thing I don't know about yet?)

timj: Thank you. With so many changes it's easy to miss some. futurize knows that `types.StringType` is a synonym for `bytes` and since `StringType` does not exist on python3 it just buts `bytes` in there.

timj: and of course this fixes my final test failure on Python 3.

python/lsst/daf/persistence/repository.py
```diff
@@ -120,7 +120,7 @@ def _initMapper(self, repositoryCfg):
                     if mro[-1] is object:
                         mro = mro[:-1]
                     for c in mro:
-                        if arg in inspect.getfullargspec(c.__init__ ).args:
+                        if arg in inspect.getargspec(c.__init__ ).args:
```
n8pease: I guess this could get squashed out? (it doesn't matter much to me personally. I just noticed it)

timj: Yes, I will squash that. It's still a problem we need to fix but I'm not entirely sure how to fix it in the short term without a python version check. All that happens at the moment is you get a deprecation warning when you run any code.

tests/repoFindByDate.py
```diff
@@ -158,7 +158,7 @@ def map_cfg(self, dataId, write):
         lookups = self.storage.lookup(lookupProperties='date', reference=None,
                                      dataId=dataId, template=template)
         lookups.sort()
-        if len(lookups) is not 0:
+        if len(lookups):
```
n8pease: IMO 'is not 0' is more expressive of intent; I'm not interested if the value is True (or False), but that the length of the list is > 0. 
I suppose there's something contrary to my perspective in python coding standards, but that's my $0.02.

timj: It should be `> 0` then. `is not` is not really the operator you are looking for. An alternative could be `if lookups` which will be true if the list contains elements. I realize that `0` is a singleton in Python but that's not generally how tests against numbers are done.

tests/DbStorage_1.py
```diff
@@ -86,7 +86,7 @@ def testWriteRead(self):
 
         db.query()
 
-        self.assertTrue(next(db))
+        self.assertTrue(db.next())
```
n8pease: if it's expected in our code that next(x) would be used instead of x.next() I think it is worth copying what you put in the checkin comment into a comment above this line so future developers & reviewers don't flag this and go thru the same process of discovery.

timj: If `db` was meant to be iterable in the Python sense then we should add the `__iter__` method to `DbStorage` rather than trying to document our way out of it. I'm not entirely sure why that's missing from the class as it is found in others. I don't know where I would add the method.

tests/repoFindByDate.py
n8pease: > I realize that 0 is a singleton in Python but that's not generally how tests against numbers are done.


oh right, I stumble on that sometimes; the `is not` operator is incorrect even if it does work. I think `len(lookups) != 0` is probably the most expressive. `if lookups` is probably the most pythonic(?)
I'm ok with either, but I like the former better.

PR #24, opened 2016-08-24
make Policy init via duck typing

n8pease:
removes kwarg style input arguments

### DIFF
--- a/python/lsst/daf/persistence/policy.py
+++ b/python/lsst/daf/persistence/policy.py
@@ -69,25 +69,31 @@ class Policy(_PolicyBase):
     - pex policy: read is supported, although this is deprecated and will at some point be removed.
     """
 
-    def __init__(self, other=None, preference=('data', 'policy', 'pexPolicy', 'filePath', 'defaultInitData'),
-                 **kwargs):
-        """Constructor. Accepts an optional vars, one of which will be used to initalize the Policy.
-        If more than initializer var is populated, one will be selected in order specified by 'preference'.
-
-        :param preference: a tuple in order of preference to be used if more than 1 initialzer var is
-                           populated.
-        :param kwargs:
-            data: a dict with initial key-value pairs.
-            policy: an instance of a Policy that should be copied into this Policy.
-            filePath: path to a policy file. May be .yaml or .paf
-            pexPolicy: an instance of a pex policy to copy into this Policy
-            defaultInitData: policy data with a populated instance of a defaultInitData
-                             e.g. Policy.defaultInitData('daf_persistence', 'testPolicy.yaml', 'tests')
-
-        :return:
+    def __init__(self, other=None):
+        """Initialize the Policy. Other can be used to initialize the Policy in a variety of ways:
+        other (string) Treated as a path to a policy file on disk. Must end with '.paf' or '.yaml'.
+        other (Pex Policy) Initializes this Policy with the values in the passed-in Pex Policy.
+        other (Policy) Copies the other Policy's values into this one.
+        other (dict) Copies the values from the dict into this Policy.
         """
         collections.UserDict.__init__(self)
-        self.__init(other, preference, **kwargs)
+
+        if other is None:
+            return
+
+        if isinstance(other, collections.Mapping):
+            self.update(other)
+        elif isinstance(other, Policy):
+            self.data = copy.deepcopy(other.data)
+        elif isinstance(other, basestring):
+            # if other is a string, assume it is a file path.
+            self.__initFromFile(other)
+        elif isinstance(other, pexPolicy.Policy):
+            # if other is an instance of a Pex Policy, load it accordingly.
+            self.__initFromPexPolicy(other)
+        else:
+            # if the policy specified by other could not be loaded raise a runtime error.
+            raise RuntimeError("A Policy could not be loaded from other:%s" % other)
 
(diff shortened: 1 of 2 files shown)


tests/testPolicy.py
```diff
@@ -111,19 +111,23 @@ def testMerge(self):
...
-        loadedPolicy = Policy(filePath='testDumpFile.yaml')
+        loadedPolicy = Policy('testDumpFile.yaml')
         self.assertEqual(self.policy, loadedPolicy)
         os.remove('testDumpFile.yaml')
 
     def testNonExistantPolicyAtPath(self):
-        self.assertRaises(IOError, Policy, filePath="c:/policy.yaml")
-        self.assertRaises(IOError, Policy, filePath="c:/policy.paf")
-        self.assertRaises(IOError, Policy, filePath="c:/policy")
+        self.assertRaises(IOError, Policy, "c:/policy.yaml")
```
timj: Can you rewrite these using `with` so it's clearer what the test is.

timj: It's a shame `FileNotFoundError` seems to be Python 3 only.

python/lsst/daf/persistence/policy.py
```diff
@@ -250,6 +210,24 @@ def __contains__(self, key):
...
+    def defaultPolicyFile(productName, filePath, repos=None):
+        """Get the path to a default policy file
+
+        @param productName (string) The name of the product that the default policy is installed as part of
+        @param filepath (string) The relative pathname to the policy file.  
+        @param repos (string) the subdirectory with the product's install directory where policy files are 
+                              stored. If None (default), the filepath argument is relative to the installation
+                              directory.
+        """
+        basePath = os.getenv(productName.upper() + "_DIR")
```
timj: Why use this instead of the `getPackageDir` that was in place before?

python/lsst/daf/persistence/policy.py
```diff
@@ -69,25 +69,36 @@ class Policy(_PolicyBase):
...
+    
         """
         collections.UserDict.__init__(self)
-        self.__init(other, preference, **kwargs)
+        
+        if other is None:
+            return
+
+        try:
+            other.keys() # is it a dict?
```
timj: The spacing to the comment is wrong here. Did you check the code was compliant (with flake8)?

timj: Also, why not just ask if it's dict-like and fit it in with the rest of the tests below?

python/lsst/daf/persistence/policy.py
```diff
@@ -250,6 +210,24 @@ def __contains__(self, key):
                 return False
         return keys[-1] in d
 
+    @staticmethod
+    def defaultPolicyFile(productName, filePath, repos=None):
+        """Get the path to a default policy file
+
+        @param productName (string) The name of the product that the default policy is installed as part of
+        @param filepath (string) The relative pathname to the policy file.
```
timj: "The location of the file relative to the product root directory."?

python/lsst/daf/persistence/policy.py
```diff
@@ -250,6 +210,24 @@ def __contains__(self, key):
...
 
+    @staticmethod
+    def defaultPolicyFile(productName, filePath, repos=None):
+        """Get the path to a default policy file
+
+        @param productName (string) The name of the product that the default policy is installed as part of
+        @param filepath (string) The relative pathname to the policy file.  
+        @param repos (string) the subdirectory with the product's install directory where policy files are 
+                              stored. If None (default), the filepath argument is relative to the installation
+                              directory.
```
timj: Ok, I'm actually having a hard time working out how `filepath` and `repos` relate to each other from this description without looking at the code. Is the actual file in `productDir/repos/filepath`? Is it also `productDir/filepath` if `repos` is `None`? Does `filepath` include the actual file itself or just the path to the file? If it includes the file does it include the extension? All these questions were in my head as I read this description.

python/lsst/daf/persistence/policy.py
n8pease: this got cut & pasted from a code I wrote before I knew about getPackageDir. will fix.

python/lsst/daf/persistence/policy.py
n8pease: This reimplements (in python) code from pex_policy. The documentation is was moved from there. You're right though, it's hard to parse. I'll clean it up.

tests/testPolicy.py
```diff
@@ -111,19 +111,25 @@ def testMerge(self):
...
+        loadedPolicy = Policy('testDumpFile.yaml')
         self.assertEqual(self.policy, loadedPolicy)
         os.remove('testDumpFile.yaml')
 
     def testNonExistantPolicyAtPath(self):
-        self.assertRaises(IOError, Policy, filePath="c:/policy.yaml")
-        self.assertRaises(IOError, Policy, filePath="c:/policy.paf")
-        self.assertRaises(IOError, Policy, filePath="c:/policy")
+        # Test that loading paf & yaml files that do not exist raises an IOError.
+        self.assertRaises(IOError, Policy, "c:/nonExistantPath/policy.yaml")
```
timj: Please use
```python
with self.assertRaises(IOError):
    Policy("c:....yaml")
```
as it is much clearer how the command is constructed that triggers the exception.

PR #27, opened 2016-09-21
Tickets/dm 7469

n8pease:


python/lsst/daf/persistence/butler.py
```diff
@@ -634,9 +642,16 @@ def put(self, obj, datasetType, dataId={}, doBackup=False, **rest):
         for repoData in self._repos.outputs():
             location = repoData.repo.map(datasetType, dataId, write=True)
             if location:
-                if doBackup:
-                    repoData.repo.backup(datasetType, dataId)
-                repoData.repo.write(location, obj)
+                if isinstance(location, ButlerComposite):
+                    components = {}
+                    location.disassembler(obj=obj, dataId=location.dataId, componentDict=components)
```
ktlim: Might be nice to prefill `components` with the names of the datasets expected and even to pass the disassembler the expected dataset types.  It's also a little confusing to have a local `components` dict as well as `location.components`.

python/lsst/daf/persistence/butler.py
```diff
@@ -606,6 +606,14 @@ def get(self, datasetType, dataId=None, immediate=False, **rest):
             bypassFunc = getattr(location.mapper, "bypass_" + datasetType)
             callback = lambda: bypassFunc(datasetType, pythonType, location, dataId)
         else:
+            if isinstance(location, ButlerComposite):
+                for name, info in location.components.items():
+                    location.components[name] = self.get(info.datasetType, location.dataId, immediate=True, 
+                                                         **rest)
```
ktlim: `**rest` has already been added to the dataId, so it shouldn't need to be passed through here.

python/lsst/daf/persistence/butlerLocation.py
```diff
@@ -27,9 +27,54 @@
...
 import lsst.daf.base as dafBase
 
+from collections import namedtuple
 import yaml
 
-from . import iterify
+from . import iterify, doImport
+
+class ButlerComposite(object):
+    componentInfo = namedtuple('componentInfo', 'datasetType subset inputOnly')
```
ktlim: This is a class, so it should be capitalized, I would think.

python/lsst/daf/persistence/butlerLocation.py
```diff
@@ -27,9 +27,54 @@
...
+        :param assembler: function object or importable string to a function object that can be called with 
+                          the assembler signature
+        :param disassembler: function object or importable string to a function object that can be called with 
+                             the assembler signature
+        :param python: a python class object or importable string to a class object that can be used by the 
+                       assembler to instantiate an object to be returned
+        :param dataId: the dataId that is used to look up components
+        :param mapper: a reference to the mapper that created this ButlerComposite object
+        """
+        self.assembler = doImport(assembler) if isinstance(assembler, basestring) else assembler
```
ktlim: Do you need a `from past.builtins import basestring` to run under Python3?

python/lsst/daf/persistence/repositoryCfg.py
```diff
@@ -27,7 +27,7 @@
 import inspect
```
ktlim: Not using `inspect` anymore, I think.

python/lsst/daf/persistence/repositoryCfg.py
```diff
@@ -123,20 +123,16 @@ def makeFromArgs(repositoryArgs, parents):
...
-            if inspect.isclass(self._mapper):
-                if not inspect.isclass(repositoryArgs.mapper):
-                    return False
-                if self._mapper != repositoryArgs.mapper:
-                    return False
-            else:
-                if type(self._mapper) != type(repositoryArgs.mapper):
-                    return False
-        # check mapperArgs for any keys in common and if their value does not match then return false.
-        if self._mapperArgs is not None and repositoryArgs.mapperArgs is not None:
```
ktlim: You don't still need to check the `mapperArgs`?

python/lsst/daf/persistence/repositoryCfg.py
```diff
@@ -123,20 +123,16 @@ def makeFromArgs(repositoryArgs, parents):
     def matchesArgs(self, repositoryArgs):
```
ktlim: docstring?

python/lsst/daf/persistence/butler.py
```diff
@@ -606,6 +606,14 @@ def get(self, datasetType, dataId=None, immediate=False, **rest):
             bypassFunc = getattr(location.mapper, "bypass_" + datasetType)
             callback = lambda: bypassFunc(datasetType, pythonType, location, dataId)
         else:
+            if isinstance(location, ButlerComposite):
+                for name, info in location.components.items():
+                    location.components[name] = self.get(info.datasetType, location.dataId, immediate=True,
```
ktlim: If this is a `subset`, this will have to be more complicated, no?

python/lsst/daf/persistence/butlerLocation.py
```diff
@@ -27,9 +27,54 @@
...
 import yaml
 
-from . import iterify
+from . import iterify, doImport
+
+class ButlerComposite(object):
+    componentInfo = namedtuple('componentInfo', 'datasetType subset inputOnly')
+
+    def __init__(self, assembler, disassembler, python, dataId, mapper):
+        """Initializer
```
ktlim: Not required now, but you might think of starting to convert to Numpydoc (https://developer.lsst.io/docs/py_docs.html#documenting-python-apis)

python/lsst/daf/persistence/__init__.py
```diff
@@ -22,6 +22,10 @@
 
 """Python interface to lsst::daf::persistence classes
 """
+import pkgutil
+import lsstimport
+__path__ = pkgutil.extend_path(__path__, __name__)
+
```
TallJimbo: Why is this needed?  I'm concerned it may be fragile to have an `__init__.py` file that has both the above invocation and any real content (such as the imports below).  It implies that there is another `__init__.py` with the same Python name (i.e. `lsst.daf.persistence`) in another location, and I believe only one of those is ever executed.

python/lsst/daf/persistence/butlerLocation.py
```diff
@@ -27,9 +27,65 @@
...
+            Function object or importable string to a function object that can be called with the disassembler
+            signature: (object, dataId, componentDict).
+        python : class object
+            A python class object or importable string to a class object that can be used by the assembler to
+            instantiate an object to be returned.
+        dataId : dict or DataId
+            The dataId that is used to look up components.
+        mapper : Mapper instance
+            A reference to the mapper that created this ButlerComposite object.
+        """
```
TallJimbo: I think docstrings for `__init__` functions don't actually work as you might expect in Python (they don't appear when you ask for on-line help on the type).  It's standard to instead document the constructor and its parameters in the class docstring.  More on that [here](https://developer.lsst.io/docs/py_docs.html#py-docstring-class-structure).

python/lsst/daf/persistence/test/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+
+"""Python interface to lsst::daf::persistence classes
+"""
```
TallJimbo: Docstring is inaccurate (and perhaps not useful here).

python/lsst/daf/persistence/test/testObject.py
```diff
@@ -0,0 +1,58 @@
...
+class TestObject(object):
+    """A basic object for testing. Contains a data blob that can be compared with other TestObject instances
+    for equality (or inequality).
+    """
+
+    def __init__(self, data):
+        self.data = data
+
+    def __eq__(self, other):
+        return self.data == other.data
```
TallJimbo: Even if you don't use it, I'd recommend implementing `__ne__` as well, just so avoid confusing behavior down the road (if you don't implement it, `!=` will fall back to pointer equality instead of the inverse of your custom equality operator).

PR #28, opened 2016-09-29
Tickets/dm 7719

n8pease:


python/lsst/daf/persistence/butler.py
```diff
@@ -195,37 +201,47 @@ class Butler(object):
...
+    Initialization:
+
+    The preferred method of initialization is to pass in a RepositoryArgs instance, or a list of
+    RepositoryArgs to inputs and/or outputs.
+
+    For backward compatibility: this initialization method signature can take a posix root path, and
+    optionally a mapper class instance or class type that will be instantiated using the mapperArgs input
+    argument. However, for this to work in a backward compatible way it creates a single repository that is
+    used as both an input and an output repository. This is NOT preferred, and will likely break any
+    provenance system we have in place.
```
TallJimbo: If we plan to remove the current approach to creating an input-only `Butler` in a single line, I think we'll need to add a new one.  Creating a butler that points to a posix-storage repo at a particular filesystem location with minimal boilerplate is absolutely a requirement (though I don't care if it's via the constructor or a convenience function).

PR #29, opened 2016-10-13
Tickets/dm 7719

n8pease:


n8pease:
I'm still checking these. @TallJimbo has checked the unit tests. I'll add a reviewer for the code when I'm ready.

python/lsst/daf/persistence/butler.py
```diff
@@ -567,6 +567,64 @@ def datasetExists(self, datasetType, dataId={}, **rest):
...
+                continue
+            location.datasetType = datasetType # todo is there a better way than monkey patching here?
+            if len(components) > 0:
+                if not isinstance(location, ButlerComposite):
+                    raise RuntimeError("todo msg; location for a dotted datasetType must be a composite.")
+                # replace the first component name with the datasetType
+                components[0] = location.componentInfo[components[0]].datasetType
+                # join components back into a dot-delimited string
+                datasetType = '.'.join(components)
+                location = self._locate(datasetType, dataId, write)
```
ktlim: I don't think this recursion does what you want, since only `components[0]` is ever modified.  That also suggests that you may need to do a test of a datasetType like "foo.bar.baz".  Also, what if the component datasetType contains one or more dots?

ktlim: Oops.  Never mind about the first part, I missed the `components = components[1:]` above.  And I think the second part might also not be an issue.

PR #32, opened 2016-10-18
Tickets/dm 7791

n8pease:


python/lsst/daf/persistence/butler.py
```diff
@@ -822,11 +828,50 @@ def dataRef(self, datasetType, level=None, dataId={}, **rest):
...
-        self.log.debug("Starting read from %s", location)
-        results = location.repository.read(location)
-        if len(results) == 1:
-            results = results[0]
-        self.log.debug("Ending read from %s", location)
+        """Unpersist an object using data inside a butlerLocation object.
+
+        A weakref to loaded objects is cached here. If the object specified by the butlerLocaiton has been
+        loaded before and still exists then the object will not be re-read. A ref to the already-existing
+        object will be reutrned instead.
```
TallJimbo: typo: `reutrned`.

python/lsst/daf/persistence/butler.py
```diff
@@ -822,11 +828,50 @@ def dataRef(self, datasetType, level=None, dataId={}, **rest):
...
+            except AttributeError: # if it's not,
+                usedDataId = butlerLocation.usedDataId # assume it's a raw dict.
+            return hash((butlerLocation.pythonType, butlerLocation.cppType, butlerLocation.storageName,
+                         id(butlerLocation.mapper), id(butlerLocation.storage),
+                         tuple(butlerLocation.locationList), json.dumps(usedDataId),
+                         butlerLocation.datasetType))
+
+        locationHash = hasher(location)
+        if locationHash in self.objectCache:
+            results = self.objectCache[locationHash]
```
TallJimbo: I'm sure it doesn't matter in this case, but since code is sometimes copied, a more efficient pattern (only one dict lookup) for this is:
```
results = self.objectCache.get(locationHash, None)
if results is not None:
    ...
```

python/lsst/daf/persistence/butler.py
```diff
@@ -822,11 +828,50 @@ def dataRef(self, datasetType, level=None, dataId={}, **rest):
...
+        def hasher(butlerLocation):
+            """Hash a butler location for use as a key in the object cache."""
+            try: # is the dataId a daf_persistence.Policy?
+                usedDataId = butlerLocation.usedDataId.data # if so, fetch the dict
+            except AttributeError: # if it's not,
+                usedDataId = butlerLocation.usedDataId # assume it's a raw dict.
+            return hash((butlerLocation.pythonType, butlerLocation.cppType, butlerLocation.storageName,
+                         id(butlerLocation.mapper), id(butlerLocation.storage),
+                         tuple(butlerLocation.locationList), json.dumps(usedDataId),
+                         butlerLocation.datasetType))
```
TallJimbo: It might be worthwhile to move this to a `ButlerLocation` method so it can be used elsewhere.  I think we may want to cache objects in `put` as well, for instance (I can definitely think of use cases for that, though I haven't thought hard about any problems it might create).

python/lsst/daf/persistence/butler.py
```diff
@@ -822,11 +828,48 @@ def dataRef(self, datasetType, level=None, dataId={}, **rest):
...
+            be fetched from the object cache (if it has already been read before).
+        """
+        def hasher(butlerLocation):
+            """Hash a butler location for use as a key in the object cache."""
+            try: # is the dataId a daf_persistence.Policy?
+                usedDataId = butlerLocation.usedDataId.data # if so, fetch the dict
+            except AttributeError: # if it's not,
+                usedDataId = butlerLocation.usedDataId # assume it's a raw dict.
+            return hash((butlerLocation.storageName, id(butlerLocation.mapper), id(butlerLocation.storage),
+                         tuple(butlerLocation.locationList), json.dumps(usedDataId),
```
brianv0: I'm not 100% sure that you'll get reproducible hashes if you do this with Python 3 unless usedDataId is an OrderedDict or something. Maybe @timj can comment.

python/lsst/daf/persistence/butler.py
```diff
@@ -822,11 +828,48 @@ def dataRef(self, datasetType, level=None, dataId={}, **rest):
...
+
+        Returns
+        -------
+        object - an instance of the object specified by the butlerLoction.
+            The object specified by the butlerLocation will either be loaded from persistent storage or will
+            be fetched from the object cache (if it has already been read before).
+        """
+        def hasher(butlerLocation):
+            """Hash a butler location for use as a key in the object cache."""
+            try: # is the dataId a daf_persistence.Policy?
```
ktlim: How does usedDataId (which looks to be documented as a dict and created as a new dict each time) get to be a Policy?

python/lsst/daf/persistence/butler.py
ktlim: Yes, I think you need to ensure the ordering of the usedDataId.  You may have to go via another tuple (in which case json may not be needed?).

python/lsst/daf/persistence/butlerLocation.py
```diff
@@ -93,7 +93,42 @@ class ButlerLocation(yaml.YAMLObject):
...
+    pythonType - string or class instance
+        This is the type of python object that should be created when reading the location.
+
+    cppType - string or None
+        The type of cpp object represented by the location (optional, may be None)
+
+    storageName - string
+        The type of storage the object is in or should be place into.
+
+    locationList - list of string
```
ktlim: With the new component infrastructure, it seems to me that the previous uses of multiple entries in locationList are now obsolete and it could be renamed locationURI.

PR #37, opened 2016-10-27
fix unittests for pytest

n8pease:


tests/DbAuth.py
```diff
@@ -38,12 +38,11 @@ class DbAuthTestCase(unittest.TestCase):
     """A test case for DbAuth."""
 
     def setUp(self):
-        self.pol = Policy(os.path.join(ROOT, "testDbAuth.paf"))
```
n8pease: it was nonsensical to keep self.pol here; DbAuth has a static shared ptr where it gets kept, and calling reset releases/deletes it.

timj: That was me misunderstanding things when I tried to fix a leak in 26518a12bda179b6b86a2afd7bd1b7e285be3fed

PR #39, opened 2016-11-09
allow dicts in sequencify

n8pease:


tests/testUtils.py
```diff
@@ -0,0 +1,54 @@
...
+    # Many of the functions in lsst.daf.persistence.utils are not represented here.
+    # https://jira.lsstcorp.org/browse/DM-8236 was created to finish the tests in this file.
+
+    def testSequencify(self):
+        self.assertEqual((1, ), dp.sequencify(1))
+        self.assertEqual(('abc', ), dp.sequencify('abc'))
+        self.assertEqual((1, ), dp.sequencify((1,)))
+        self.assertEqual([1], dp.sequencify([1]))
+        self.assertEqual(set([1]), dp.sequencify(set([1])))
+        self.assertEqual(('a', 'b', 'c'), dp.sequencify({'a':1, 'b':2, 'c':3}))
```
timj: Can you add a test that `('b', 'c', 'a')` does not `assertEqual`? So that we know that we aren't getting a match by chance.

PR #41, opened 2016-11-22
daf_persistence - Tickets/dm 7979

n8pease:


python/lsst/daf/persistence/posixStorage.py
```diff
@@ -105,6 +105,10 @@ def putRepositoryCfg(cfg, loc=None):
             cfg = copy.copy(cfg)
             loc = cfg.root
             cfg.root = None
+        # This class supports schema 'file' and also treats no schema as 'file'.
+        # Split the URI and take only the path; remove the schema fom loc if it's there.
+        parseRes = urllib.parse.urlparse(loc)
```
brianv0: Do you want to check that parseRes.scheme == "file"?

n8pease: it gets checked in storage.py, I decided to not do redundant verification here.

PR #45, opened 2017-02-18
DM-9112: Determine strategy for dealing with butler proxies in pybind11

pschella:


include/lsst/daf/persistence/pybind11/readProxy.h
```diff
@@ -0,0 +1,79 @@
...
+};
+
+/** Register a type for (optional) lazy load.
+  * 
+  * When called in the pybind11 wrapper this function adds an implict conversion
+  * from ReadProxy to the registered type.
+  * 
+  * @tparam OutputType Type to register (e.g. DateTime)
+  */
+
```
TallJimbo: Superfluous newline (which may or may not mess up Doxygen parsing, but best to be safe and remove it).

include/lsst/daf/persistence/pybind11/readProxy.h
```diff
@@ -0,0 +1,79 @@
...
+template <typename OutputType> void register_proxy() {
+    namespace py = pybind11;
+    auto implicit_caster = [](PyObject *obj, PyTypeObject *type) -> PyObject * {
+        if (!py::detail::make_caster<lsst::daf::persistence::python::ReadProxyBase>().load(obj, false))
+            return nullptr;
```
TallJimbo: Please use braces around `if` block.

include/lsst/daf/persistence/pybind11/readProxy.h
```diff
@@ -0,0 +1,79 @@
...
+        if (result == nullptr) {
+            PyErr_Clear();  // needed to fall through to next conversion
+        }
+        return result;
+    };
+
+    if (auto tinfo = py::detail::get_type_info(typeid(OutputType)))
+        tinfo->implicit_conversions.push_back(implicit_caster);
+    else
+        py::pybind11_fail("register_proxy: Unable to find type " + py::type_id<OutputType>());
```
TallJimbo: Please add braces.

Also, I assume this failure occurs when `register_proxy<T>()` is called before `py::class_<T,...>` is constructed?  If so, it might be helpful to include that in the error message.

python/lsst/daf/persistence/persistence.cc
```diff
@@ -7,24 +7,32 @@
...
+namespace daf {
+namespace persistence {
 
 PYBIND11_PLUGIN(_persistence) {
     py::module mod("_persistence", "Access to the classes from the daf_persistence persistence library");
 
-    py::class_<Persistence, std::shared_ptr<Persistence>, lsst::daf::base::Citizen> cls(mod, "Persistence");
+    py::class_<python::ReadProxyBase, std::shared_ptr<python::ReadProxyBase>>(mod, "ReadProxyBase")
+        .def(py::init<>())
+        .def_readwrite("subject", &python::ReadProxyBase::subject);
```
TallJimbo: Could we call this `__subject__` and remove `__subject__` from `ReadProxy.__slots__`?  It's possible we might also have to replace `object.__getattribute__` in `ReadProxy` with `ReadProxyBase.__getattribute__` (I don't know if pybind11 overrides that, but it's conceivable that it does for e.g. static properties).

In any case, I do think it'd be worth some effort to avoid having the proxy introduce another attribute, *especially* a non-dunder one.

tests/butlerProxy.py
```diff
@@ -0,0 +1,81 @@
...
+        # First try with immediate read, this should obviously work
+        dt = self.butler.get("dt", ccd=1, immediate=True)
+        self.assertIsInstance(dt, dafBase.DateTime)
+        self.assertTrue(isValidDateTime(dt))
+
+        # Now try again with lazy read
+        dt = self.butler.get("dt", ccd=1, immediate=False)
+        self.assertIsInstance(dt, dafPersist.readProxy.ReadProxy)
+        self.assertTrue(isValidDateTime(dt))
+
```
TallJimbo: It might be good to try passing something that isn't a valid `DateTime` or a proxy to `isValidDateTime` to make sure the exception type (at least) and hopefully the message are what they would be if we didn't register a proxy for it.

tests/testLib.cc
```diff
@@ -0,0 +1,24 @@
+#include <pybind11/pybind11.h>
```
TallJimbo: Use quotes, not angle brackets for this include.

include/lsst/daf/persistence/pybind11/readProxy.h
```diff
@@ -0,0 +1,79 @@
...
+ * GNU General Public License for more details.
+ * 
+ * You should have received a copy of the LSST License Statement and 
+ * the GNU General Public License along with this program.  If not, 
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+ 
+#ifndef LSST_DAF_PERSISTENCE_PYTHON_READPROXY_H
+#define LSST_DAF_PERSISTENCE_PYTHON_READPROXY_H
+
```
TallJimbo: Missing some includes?  Needs at least `#include "pybind11/pybind11.h"`

PR #50, opened 2017-03-17
daf_persistence add functions to query if a storage exists at a location

n8pease:


python/lsst/daf/persistence/posixStorage.py
```diff
@@ -601,6 +606,22 @@ def search(root, path, searchParents=False):
...
+        ----------
+        root : string
+            URI to the the root location of the storage
+
+        Returns
+        -------
+        bool
+            True if the storage exists, false if not
+        """
+        return os.path.exists(PosixStorage._pathFromURI(uri))
```
jhoblitt: It looks like this boils down to a `stat(2)` call, so this should be nice and atomic.

jhoblitt:
This looks reasonable to me but I am unfamiliar with this code.

python/lsst/daf/persistence/posixStorage.py
```diff
@@ -50,7 +50,7 @@ def __init__(self, uri):
         :return:
         """
         self.log = Log.getLogger("daf.persistence.butler")
-        self.root = urllib.parse.urlparse(uri).path
+        self.root = self._pathFromURI(uri)
         if self.root and not os.path.exists(self.root):
             os.makedirs(self.root)
```
timj: I think this should use the safe `safeMakeDir()`.

python/lsst/daf/persistence/posixStorage.py
```diff
@@ -601,6 +606,22 @@ def search(root, path, searchParents=False):
             else:
                 return None
 
+    @staticmethod
+    def storageExists(uri):
```
timj: This method is added but nothing in this commit seems to call it. Is it a specialization of a base implementation?

python/lsst/daf/persistence/posixStorage.py
jhoblitt: I missed this... is using the `exist_ok` param not enough?

python/lsst/daf/persistence/posixStorage.py
n8pease: it's called by the new code in storageContinued.py

python/lsst/daf/persistence/posixStorage.py
jhoblitt: I looked at `safeMakeDir` and I'm not sure if it is actually safe if the path it is given contains multiple [new] nested dirs.

python/lsst/daf/persistence/posixStorage.py
jhoblitt: I believe @timj's concern is valid.  Two threads could both make the `stat(2)` call while the dir does not exist and then collide on directory creation.

PR #51, opened 2017-03-20
daf_persistence Tickets/dm 7468

n8pease:


python/lsst/daf/persistence/storageInterface.py
```diff
@@ -0,0 +1,272 @@
...
+            The location that was found, or None if no location was found.
+        """
+
+    @abstractmethod
+    def copyFile(self, fromLocation, toLocation):
+        """Copy a file from one location to another on the local filesystem.
+
+        Parameters
+        ----------
+        fromLocation : path
```
andy-slac: is `path` a `str`? I think we are supposed to list actual Python types on the right side.

python/lsst/daf/persistence/storageInterface.py
```diff
@@ -0,0 +1,272 @@
...
+        -------
+        None
+        """
+
+    @abstractmethod
+    def locationWithRoot(self, location):
+        """Get the full path to the location.
+
+        :param location:
+        :return:
```
andy-slac: Different style of docstring

python/lsst/daf/persistence/storageInterface.py
```diff
@@ -0,0 +1,272 @@
...
+        """
+
+    @staticmethod
```
andy-slac: I'd probably made all `staticmethod`s `classmethod`s, latter works nicer with inheritance.

python/lsst/daf/persistence/storageInterface.py
```diff
@@ -0,0 +1,272 @@
...
+        -------
+        A RepositoryCfg instance or None
+        """
+
+    @staticmethod
+    @abstractmethod
+    def putRepositoryCfg(cfg, loc=None):
+        """Serialize a RepositoryCfg to a location.
+
+        When loc == cfg.root, the RepositoryCfg is to be writtenat the root
```
andy-slac: "writtenat" -> "written at"

python/lsst/daf/persistence/storageInterface.py
```diff
@@ -0,0 +1,272 @@
...
+        location of the repository. In that case, root is not written, it is
+        implicit in the location of the cfg. This allows the cfg to move from
+        machine to machine without modification.
+
+        Parameters
+        ----------
+        cfg : RepositoryCfg instance
+            The RepositoryCfg to be serailized.
+        loc : None, optional
```
andy-slac: What is the type if it's not `None`?

python/lsst/daf/persistence/storageInterface.py
```diff
@@ -0,0 +1,272 @@
...
+
+    @staticmethod
+    @abstractmethod
+    def getMapperClass(root):
+        """Get the mapper class associated with a repository root.
+
+        Parameters
+        ----------
+        root : string
+            The location of a persisted ReositoryCfg is (new style repos).
```
andy-slac: "ReositoryCfg" -> "RepositoryCfg"

python/lsst/daf/persistence/swiftStorage.py
```diff
@@ -0,0 +1,519 @@
...
+
+
+class SwiftStorage(StorageInterface):
+    """Storage Interface implementation specific for Swift
+
+    Requires that the following environment variables exist:
+    SWIFT_USERNAME : string
+        The username to use when authorizing the connection.
+    SWIFT_PASSWORD : string
+        The password to use when authorizing the connection.
```
andy-slac: Using envvar for secrets is somewhat (or very) unsecure, will it be replaced by something else in the future?

python/lsst/daf/persistence/swiftStorage.py
```diff
@@ -0,0 +1,519 @@
...
+        swiftLocations = butlerLocation.getLocations()
+        # Here the ButlerLocation is modified sligtly to write to a temporary
+        # file via PosixStorage. (Then the temporary file is written to the
+        # swift container)
+        localFile = tempfile.NamedTemporaryFile()
+        butlerLocation.locationList = [localFile.name]
+        butlerLocation.storage = PosixStorage('/')
+        butlerLocation.storage.write(butlerLocation, obj)
+        butlerLocation.locationList = swiftLocations
+        with open(localFile.name, 'rb') as f:
```
andy-slac: Can you avoid re-opening the file? If think `tempfile` warns that opening temporary file second time it is non-portable (and there are other reasons to avoid that). I think you can just do `localFile.seek(0)` and then read everything from it.

python/lsst/daf/persistence/swiftStorage.py
```diff
@@ -0,0 +1,519 @@
...
+        Returns
+        -------
+        A list of objects as described by the butler location. One item for
+        each location in butlerLocation.getLocations()
+        """
+        localFile = self.getLocalFile(butlerLocation.getLocations()[0])
+        butlerLocation.locationList = [localFile.name]
+        butlerLocation.storage = PosixStorage('/')
+        obj = butlerLocation.storage.read(butlerLocation)
+        return obj
```
andy-slac: The implementation is not consistent with docstrings which says that a list is returned and it includes all locations, not just first one?

python/lsst/daf/persistence/swiftStorage.py
```diff
@@ -0,0 +1,519 @@
...
+        ------
+        RuntimeError
+            If there is an error downloading the object.
+        """
+        self._log.info("SwiftStorage getting file {}".format(location))
+        start = time.time()
+        try:
+            f = self._getLocalFile(location)
+            self._log.info(
+                "...getting file took {} seconds.".format(time.time() - start))
```
andy-slac: Some people may not like too much nose at INFO log level.

python/lsst/daf/persistence/swiftStorage.py
```diff
@@ -0,0 +1,519 @@
...
+            A filename (and optionally prefix path) to search for within root.
+
+        Returns
+        -------
+        string or None
+            The location that was found, or None if no location was found.
+        """
+        # Now search for the path in the root or its parents
+        # Strip off any cfitsio bracketed extension if present
+        strippedPath, pathStripped = self.getFitsHeaderStrippedPath(path)
```
andy-slac: `strippedPath, pathStripped` is not going to confuse anyone? :smile_cat:

python/lsst/daf/persistence/swiftStorage.py
```diff
@@ -0,0 +1,519 @@
...
+        try:
+            headers, objects = self._connection.get_container(
+                self._containerName)
+            locations = [obj['name'] for obj in objects]
+        except self.swift.ClientException as err:
+            raise RuntimeError("Container \'{0}\' not found: {1}".format(
+                self._containerName, err))
+        import fnmatch
```
andy-slac: Should probably go to the top of the file.

python/lsst/daf/persistence/swiftStorage.py
```diff
@@ -0,0 +1,519 @@
...
+        """
+        import swiftclient as swift
+        storage = SwiftStorage(uri)
+        try:
+            localFile = storage._getLocalFile('repositoryCfg')
+        except swift.ClientException:
+            # Assuming the file does not exist, not some other kind of error
+            # (how can we do more precise handling of this expcetion?)
+            return None
+        with open(localFile.name, mode='r') as f:
```
andy-slac: `localFile` is already a file object, can you read from it directly?

tests/testSwiftStorage.py
```diff
@@ -0,0 +1,199 @@
...
+import lsst.daf.persistence as dp
+import lsst.daf.persistence.test as dpTest
+from lsst.utils import getPackageDir
+import lsst.utils.tests
+import pickle
+import shutil
+import sys
+import uuid
+
+packageDir = os.path.join(getPackageDir('daf_persistence'))
```
andy-slac: Is `os.path.join` doing anything here?

python/lsst/daf/persistence/swiftStorage.py
n8pease: They tell me this is the standard way of managing credentials for swift, I don't have any plans to change this. However, the mechanism here can be extended or changed at any time if a better way is identified.

python/lsst/daf/persistence/swiftStorage.py
n8pease: Would debug be better, or what would you suggest?
For the time being I find it very useful for profiling Butler behavior, although I agree it will end up being noisy. I'd like to leave something that will print during unit test execution, but I don't need it to be 'info'.

RobertLuptonTheGood: I'd say that this was at the trace level, not even debug

python/lsst/daf/persistence/swiftStorage.py
n8pease: It confuses me! Actually it's pretty descriptive if you sit on your zen cushion for a minute. In any case, it's moved here from code before my time, and I don't want to mess with it right now.

python/lsst/daf/persistence/swiftStorage.py
andy-slac: I believe we log at INFO by default, so if you want to see this output from unit test at DEBUG/TRACE level you need to configure logging with non-default values.

python/lsst/daf/persistence/swiftStorage.py
andy-slac: OK, I'll practice zen, it'll just take longer to review if I need to focus for a minute on that line :smile_cat:

python/lsst/daf/persistence/swiftStorage.py
andy-slac: Hmm, security through full exposure of passwords. It should work nicely :scream_cat:

python/lsst/daf/persistence/swiftStorage.py
```diff
@@ -0,0 +1,555 @@
...
+        Returns
+        -------
+        string or None
+            The location that was found, or None if no location was found.
+        """
+        storage = SwiftStorage(uri, create=False)
+        return storage.instanceSearch(path) if storage else None
+
+    def copyFile(self, fromLocation, toLocation):
+        """Copy a file from one location to another on the local filesystem.
```
airnandez: From the implementation of this method I would say that the purpose here is to copy one Swift object into another Swift object, therefore in the same container, not in the local filesystem. If that's the intention, this comment needs to be updated.

PR #52, opened 2017-03-30
DM-9918: Robustify sqlite use

PaulPrice:


python/lsst/daf/persistence/registries.py
```diff
@@ -74,7 +74,10 @@ def create(location):
         #     return CalibRegistry(location)
 
         # look for an sqlite3 registry
-        if haveSqlite3 and re.match(r'.*\.sqlite3', location):
+        if re.match(r'.*\.sqlite3', location):
+            if not haveSqlite3:
+                raise RuntimeError("sqlite3 registry specified (%s), but unable to import sqlite3 module" %
```
n8pease: this line looks longer than 80 chars (is that rule official yet?) and are we preferring .format to % string substitution now?

PaulPrice: On both points I have followed the current style in that file.

PR #57, opened 2017-05-09
DM-10267: Port HSC support for PostgreSQL registries to LSST

PaulPrice:


python/lsst/daf/persistence/registries.py
```diff
@@ -385,3 +397,78 @@ def __init__(self, location):
         else:
             conn = None
         SqlRegistry.__init__(self, conn)
+
+
+class PgsqlRegistry(SqlRegistry):
+    """A PostgreSQL-based registry"""
+    placeHolder = "%s"
```
n8pease: It seems like it would make sense to move `placeHolder = "?" ` from line 291 to the SqliteRegistry subclass. If you wanted to get really fancy about it, I think it would make sense to make SqlRegistry an abc, and create an abstract method `placeholder` that subclasses would then be forced to implement.

python/lsst/daf/persistence/registries.py
```diff
@@ -385,3 +397,78 @@ def __init__(self, location):
...
+
+    def __init__(self, location):
+        """Constructor
+
+        Parameters
+        ----------
+        location : `str`
+            Path to PostgreSQL configuration file.
+        """
```
n8pease: init method documentation is supposed to go in the class docstring I think.

PR #58, opened 2017-05-19
daf_persistence remove butler object cache

n8pease:


python/lsst/daf/persistence/butler.py
```diff
@@ -1094,37 +1088,12 @@ def _read(self, location):
         Returns
         -------
         object - an instance of the object specified by the butlerLoction.
```
andy-slac: butlerLoction -> butlerLocation (or should it be just `location`)

python/lsst/daf/persistence/butler.py
```diff
@@ -1094,37 +1088,12 @@ def _read(self, location):
...
-                self.objectCache[locationHash] = results
-            except TypeError:
-                # some object types (e.g. builtins, like list) do not support weakref, and will raise a
-                # TypeError when we try to create the weakref. This is ok, we simply will not keep those
-                # types of objects in the cache.
-                pass
+        self.log.debug("Starting read from %s", location)
+        results = location.repository.read(location)
+        if len(results) == 1:
+            results = results[0]
```
andy-slac: I remember reviewing piece of code where you did something like:

    if isinstance(results, (list, tuple)) and len(results) == 1:
        results = results[0]

to protect against cases when user type is a sequence-like object. Should not this logic apply here too?

n8pease: it's a good idea. but, in this case the the code is moved out one indent level but is otherwise the same as it ever has been, and I don't want to get into the implications of reading from a maybe-list.

RobertLuptonTheGood: Is it safer to try the indexing in a try block, and make results a list if it throws?

PR #59, opened 2017-06-02
Tickets/dm 10767

n8pease:


tests/testPosixStorage.py
```diff
@@ -139,7 +139,14 @@ def testAbsolutePath(self):
         f.close()
         del f
         f = storage.getLocalFile('foo.txt')
-        self.assertIsInstance(f, file)
+        try:
+            fileType = file
+        except NameError:
+            from io import IOBase
+            fileType = IOBase
```
r-owen: Can you use use io.IOBase in in Python 2? It is part of the standard library. If it works, I think using it would be a lot cleaner than the above. Just import it and use it.

If that does not work for some reason, then I think this would be significantly cleaner if your new code was put with the other import statements at the top of the file.

tests/butlerProxy.py
```diff
@@ -41,16 +41,17 @@
...
+        self.tearDown()
+        self.butler = dafPersist.Butler(self.inputDir,
+                                        outPath=os.path.join(self.outputDir, "proxyOut"))
 
     def tearDown(self):
-        del self.butler
-        if os.path.exists(os.path.join(ROOT, 'root/out')):
-            shutil.rmtree(os.path.join(ROOT, 'root/out'))
+        if os.path.exists(self.outputDir):
+            shutil.rmtree(self.outputDir)
```
r-owen: Is it safe to not manually delete the butler? If it holds onto any objects that inherit from `Citizen` then failure to do so will trigger an error in the memory test.

tests/testPosixStorage.py
brianv0: `file` does not inherit from `io.IOBase` in python 2.

tests/butlerProxy.py
r-owen: It seems risky to have `setUp` call `tearDown`, since normally users are free to add things to `tearDown` without worrying about breaking other methods such as `setUp`. Instead, I suggest that you make a new method `deleteOutputDir` that does what `tearDown` does above, then have `setUp` and `tearDown` call that.

Note that it would then be safe to have `tearDown` delete the butler again, which demonstrates my point.

tests/testPosixStorage.py
r-owen: Also, if you do need to still assign the class to a variable, please use a variable name that starts with a capital letter.

brianv0: +1 on putting it in the imports though

tests/testPosixStorage.py
```diff
@@ -139,7 +139,14 @@ def testAbsolutePath(self):
...
+
+        isinstance(f, fileType)
```
brianv0: You can probably get rid of this line, since the next line does the assert

tests/butlerProxy.py
n8pease: del doesn't actually do anything to destroy the object, it makes it available for garbage collection. when the test class is destroyed, the butler goes out of scope, same as if del had been called.
(and, the memory tests pass.)

tests/testPosixStorage.py
r-owen: I'm curious whether it matters that `file` inherits from `io.IOBase`. I realize that it may mean a tiny change to one test, but is that a bad thing?

tests/butlerProxy.py
r-owen: Yes the memory test passes now, but if butler ever is modified to include an object that inherits from Citizen it will stop passing and folks may be surprised at this. An explicit delete is safer.

r-owen: In any case I think refactoring the code as suggested is significantly cleaner.

n8pease: I don't see how allowing the butler instance to be GC'd will not delete a contained Citizen object whereas explicitly handing the butler to the GC will delete the contained Citizen...?

r-owen: Just to clarify: it is definitely true that we must explicitly delete test case attributes that inherit from Citizen in `tearDown`. If we don't, the memory test fails. Allowing such objects to be garbage collected definitely does not work. It is not a question of whether the objects eventually go out of scope -- of course they do -- it's whether they go out of scope fast enough that the memory test doesn't think memory has been leaked.

Also, I thought CPython did delete immediately when the ref count dropped to 0 (unless there are cycles, which take longer to clean up?) -- though I agree the language specification does not require that. If this was not the case, then I'm surprised our memory test is usable; I would expect random failures in most of our packages.

r-owen: I *suspect* that the test case instance is not deleted until after the memory test has run, so by then it's too late for automatic garbage collection to help us. (I can't think why else we'd have to explicitly delete test case instances to avoid false memory leak errors).

RobertLuptonTheGood:
I think that the test object remains in scope, so the GC doesn't happen until the process exists.

n8pease:
so @r-owen is right, and the memory test fails. It does seem to be timing or platform related (passed on centos-6, failed on centos-7).

PR #66, opened 2017-07-11
daf_persistence - Tickets/dm 11251

n8pease:


python/lsst/daf/persistence/butler.py
```diff
@@ -1511,21 +1502,35 @@ def dataRef(self, datasetType, level=None, dataId={}, **rest):
...
+                else:
+                    obj = self.get(componentInfo.datasetType, location.dataId, immediate=True)
+                    componentInfo.obj = obj
+                assembler = location.assembler or genericAssembler
+            results = assembler(dataId=location.dataId, componentInfo=location.componentInfo, cls=location.python)
+            return results
+        else:
+            results = location.repository.read(location)
+            if len(results) == 1:
+                results = results[0]
```
kennywlo: The above assignment is changing the type for 'results' on the fly - consider using a new variable?

PR #70, opened 2017-07-25
Tickets/dm 11284

n8pease:


tests/test_repository.py
```diff
@@ -840,6 +815,135 @@ def makeRegistry(location, name):
...
+                               outputs=repoBDir)
+        # initializing the butler, but only specifiying the root of 'repoB' (the rest of the cfg will be
+        # loaded from its `repositoryCfg` file), but fully specify the input as required.
+        butler = dp.Butler(inputs={'root': self.repoADir,
+                                   'mapperArgs': {'foo': 1}},
+                           outputs=repoBDir)
+        self.assertEqual(butler._repos.inputs()[0].repo._mapper.kwargs, {'foo': 1})
+        # use 'repoB' as an input, and verify that 'repoA' is loaded, including the mapperArgs.
+        butler = dp.Butler(inputs=repoBDir)
+        self.assertEqual(butler._repos.inputs()[1].repo._mapper.kwargs, {'foo': 1})
```
andy-slac: I guess assumption here is that `inputs()[1]` refers to repoA and not repoB? Is that ordering checked by some other test?

tests/test_repository.py
```diff
@@ -840,6 +815,135 @@ def makeRegistry(location, name):
...
+
+
+class TestOldButlerParentTagging(unittest.TestCase):
+    """A test to verify that when a parent is an old butler repo that it still gets loaded correctly,
+    including mapperArgs and tagging."""
```
andy-slac: This is the same comment as for `TestOldButlerParent`, do they test the same thing?

python/lsst/daf/persistence/test/testMapper.py
```diff
@@ -21,11 +21,37 @@
...
+    def __init__(self, root, **kwargs):
+        self.root = root
+        self.storage = Storage.makeFromURI(self.root)
+
+    def map_foo(self, dataId, write):
+        python = TestObject
+        persistable = None
+        storage = 'PickleStorage'
+        fileName = 'filename'
+        for key, value in dataId.items():
```
andy-slac: Maybe `sorted(dataId.items())` to make sure that file name does not depend on the order of the items returned from `items()` (I think that order is not guaranteed to be identical even if content of dict is identical though maybe I'm confusing it with something else)?

python/lsst/daf/persistence/butler.py
```diff
@@ -725,23 +727,29 @@ def _getOldButlerRepositoryCfg(repositoryArgs):
...
+            and is used as the cfgRoot (URI to the location of the cfg file).
 
         Returned
         --------
         (RepositoryCfg or None, bool)
             The RepositoryCfg, or None if one cannot be found, and True if the RepositoryCfg was created by
             reading an Old Butler repository, or False if it is a New Butler Repository.
         """
+        if not isinstance(repositoryArgs, RepositoryArgs):
+            repositoryArgs = RepositoryArgs(cfgRoot=repositoryArgs, mode='r')
```
n8pease: nice catch. copy-paste error on the docs. it should be `repositoryArgs : RepositoryArgs or string`, will fix.

tests/test_repository.py
n8pease: yes, inputs & parents and repo order is covered by other tests.

python/lsst/daf/persistence/test/testMapper.py
n8pease: yeah, good catch.

PR #71, opened 2017-07-27
daf_persistence - convert mapperArgs to dict if it is None

n8pease:


tests/test_repositoryCfg.py
```diff
@@ -243,6 +243,14 @@ def test(self):
     # todo include a repositoryCfg in parents lists in this test
 
 
+class TestMapperArgsNone(unittest.TestCase):
+    """Tests that the RepositoryCfg.mapperArgs is converted to a dict if None is passed in.
+    """
+    def test(self):
+        cfg = dp.RepositoryCfg(root='foo', mapper='foo', mapperArgs=None, parents=None, policy=None)
+        self.assertIsInstance(cfg.mapperArgs, dict)
```
kennywlo: Missing _ for referencing mapperArgs.

n8pease: it's confusing from this diff; _mapperArgs is a private parameter, accessed by a property mapperArgs, which is not shown here. I discussed this with Kenny & got the sign-off.

PR #72, opened 2017-08-06
DM-7457: Use temp dirs for test repositories.

jdswinbank:


n8pease:
I've been making test output dirs unique by joining the name of the test class to the test directory. This is probably better and requires less cleanup. It might make debugging more difficult, because to examine the outputs the test output dir will have to be looked up for every execution. It may be possible to work around that, e.g. with a temporary symlink. So let's give it a shot; go ahead.

n8pease:
didn't mean to close.

PR #73, opened 2017-08-15
DM-11595: Use mkdtemp for test directories

timj:
@n8pease work in progress. In particular test_repository.TestBasics doesn't work and I don't know why...

### DIFF
--- a/tests/test_mapperImport.py
+++ b/tests/test_mapperImport.py
@@ -39,16 +39,15 @@ class MapperImportTestCase(unittest.TestCase):
     """A test case for the data butler finding a Mapper in a root"""
 
     def setUp(self):
-        self.testDir = os.path.join(ROOT, 'root')
-        self.rootPath = tempfile.mkdtemp(dir=self.testDir, prefix="out-")
-        self.outPath = os.path.split(self.rootPath)[1]
+        self.inputDir = os.path.join(ROOT, 'root')
+        self.testDir = tempfile.mkdtemp(dir=ROOT, prefix="out-")
 
-        self.butler = dafPersist.Butler(self.testDir, outPath=self.outPath)
+        self.butler = dafPersist.Butler(self.inputDir, outPath=self.testDir)
 
     def tearDown(self):
         del self.butler
-        if os.path.exists(self.rootPath):
-            shutil.rmtree(self.rootPath)
+        if os.path.exists(self.testDir):
+            shutil.rmtree(self.testDir)
 
     def testMapperClass(self):
         repository = self.butler._repos.outputs()[0].repo
@@ -59,7 +58,7 @@ class MapperImportTestCase(unittest.TestCase):
         y = butler.get("x", ccd=ccd, immediate=True)
         self.assertEqual(bbox, y)
         self.assertTrue(os.path.exists(
-            os.path.join(self.rootPath, "foo%d.pickle" % ccd)))
+            os.path.join(self.testDir, "foo%d.pickle" % ccd)))
 
     def testIO(self):
         bbox = [[3, 4], [5, 6]]
@@ -77,6 +76,7 @@ class TestMemory(lsst.utils.tests.MemoryTestCase):
 
 def setup_module(module):
     lsst.utils.tests.init()
+
 
 if __name__ == "__main__":
     lsst.utils.tests.init()
--- a/tests/test_mapperImport.py
+++ b/tests/test_mapperImport.py
@@ -25,6 +25,7 @@ import os
 import pickle
 import shutil
 import unittest
+import tempfile
 import lsst.utils.tests
 
 import lsst.daf.persistence as dafPersist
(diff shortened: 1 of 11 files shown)


tests/test_mapperImport.py
```diff
@@ -38,15 +39,16 @@ class MapperImportTestCase(unittest.TestCase):
     """A test case for the data butler finding a Mapper in a root"""
 
     def setUp(self):
-        if os.path.exists(os.path.join(ROOT, 'root/out')):
-            shutil.rmtree(os.path.join(ROOT, 'root/out'))
+        self.testDir = os.path.join(ROOT, 'root')
```
n8pease: nit: I think inputDir would be more accurate?

tests/test_mapperImport.py
```diff
@@ -38,15 +39,16 @@ class MapperImportTestCase(unittest.TestCase):
...
+        self.rootPath = tempfile.mkdtemp(dir=self.testDir, prefix="out-")
+        self.outPath = os.path.split(self.rootPath)[1]
```
n8pease: by removing the absolute part of the path and passing only the temp dir name to the outPath iarg, you're creating an output repository relative to the current working dir. Which would be fine, but in tearDown you're deleting self.rootPath, not self.outPath, so outpath will never get cleaned up. Right? I think you can just pass self.rootPath to the outPath iarg. 
Or maybe I'm faded from travel. I need to board... off I go.

tests/test_mapperImport.py
```diff
@@ -56,8 +58,8 @@ def checkIO(self, butler, bbox, ccd):
         butler.put(bbox, "x", ccd=ccd)
         y = butler.get("x", ccd=ccd, immediate=True)
         self.assertEqual(bbox, y)
-        self.assert_(os.path.exists(
-            os.path.join(ROOT, "root", "out", "foo%d.pickle" % ccd)))
+        self.assertTrue(os.path.exists(
+            os.path.join(self.rootPath, "foo%d.pickle" % ccd)))
```
n8pease: I think this will fail if the current working dir is not the same as the directory above outpath?

n8pease: confirmed; running tests from inside `daf_persistence/python` (as an example) fails
```
n8pease@Nates-MacBook-Pro ~/3/lsstsw/build/daf_persistence/python[tickets/DM-11595*]$ py.test ~/3/lsstsw/build/daf_persistence/tests/test_mapperImport.py 
```

tests/test_reposInButler.py
```diff
@@ -98,15 +99,12 @@ def map_str(self, dataId, write):
...
 class ReposInButler(unittest.TestCase):
 
-    def clean(self):
-        if os.path.exists(os.path.join(ROOT, 'repoOfRepos')):
-            shutil.rmtree(os.path.join(ROOT, 'repoOfRepos'))
-
-    def setup(self):
-        self.clean()
+    def setUp(self):
+        self.testDir = tempfile.mkdtemp(dir=ROOT, prefix='repoOfRepos-')
```
n8pease: not your bug but while your here you could fix "repoOfRepos" to accurately show the test class name "reposInButler"

tests/test_repository.py
```diff
@@ -374,10 +378,12 @@ class TestMultipleInputs(unittest.TestCase):
         - verify dissimilar datasets with same id that were written to the repos
 
     """
+    def setUp(self):
+        self.testDir = tempfile.mkdtemp(dir=ROOT, prefix="TestMapperInference-")
```
n8pease: the prefix here should be "TestMultipleInputs"

tests/test_repository.py
```diff
@@ -763,9 +768,8 @@ class TestParentRepository(unittest.TestCase):
     """A test to verify that when a parent repository is used the correct one is found."""
 
     def setUp(self):
-        self.testDir = os.path.join(ROOT, 'TestParentRepository')
-        if os.path.exists(self.testDir):
-            shutil.rmtree(self.testDir)
+        self.testDir = tempfile.mkdtemp(dir=ROOT, prefix="TestBasics-")
```
n8pease: prefix should be "TestParentRepository"

tests/test_repositoryCfg.py
```diff
@@ -39,102 +39,105 @@
 class TestCfgRelationship(unittest.TestCase):
 
     def setUp(self):
-        self.tearDown()
+        self.testDir = tempfile.mkdtemp(dir=ROOT,
+                                        prefix="repositoryCfg-")
```
n8pease: prefix should be TestCfgRelationship

tests/test_repositoryCfg.py
```diff
@@ -157,37 +160,36 @@ def testStorageRepoCfgCache(self):
...
+        self.assertEqual(cfg, storage.repositoryCfgs[os.path.join(self.testDir, 'a')])
 
 class TestNestedCfg(unittest.TestCase):
 
-    rootDir = os.path.join(ROOT, 'repositoryCfg_TestNestedCfg')
-
     def setUp(self):
-        self.tearDown()
+        self.testDir = tempfile.mkdtemp(dir=ROOT,
+                                        prefix="repositoryCfg_TestNestedCfg-")
```
n8pease: I see you just used what was already there but the prefix here can probably just be TestNestedCfg

tests/test_repositoryCfg.py
```diff
@@ -214,19 +216,18 @@ class TestCfgFileVersion(unittest.TestCase):
     """
 
     def setUp(self):
-        self.tearDown()
+        self.testDir = tempfile.mkdtemp(dir=ROOT,
+                                        prefix="repositoryCfg-")
```
n8pease: prefix -> TestCfgFileVersion

tests/test_safeFileIo.py
```diff
@@ -45,37 +46,37 @@ def setup_module(module):
 class WriteOnceCompareSameTest(unittest.TestCase):
 
     def setUp(self):
-        self.tearDown()
+        self.testDir = tempfile.mkdtemp(dir=ROOT, prefix='safeFileIo-')
```
n8pease: prefix -> "WriteOnceCompareSameTest"

tests/test_repository.py
```diff
@@ -763,9 +768,8 @@ class TestParentRepository(unittest.TestCase):
...
+        self.testDir = os.path.join(self.testDir, 'TestParentRepository')
```
n8pease: this line shouldn't be here? It's preventing files inside `TestBasics-...` (which should be `TestParentRepository-...`) 
from getting removed.

timj: Of course. Thanks for spotting that. I had to change many many lines...

PR #78, opened 2017-09-08
DM-11862: Fixes to PosixStorage read/write

n8pease:


python/lsst/daf/persistence/posixStorage.py
```diff
@@ -709,8 +711,11 @@ def writeFitsCatalogStorage(butlerLocation, obj):
     locations = butlerLocation.getLocations()
     with SafeFilename(os.path.join(butlerLocation.getStorage().root, locations[0])) as locationString:
         logLoc = LogicalLocation(locationString, additionalData)
-        flags = additionalData.getInt("flags", 0)
-        obj.writeFits(logLoc.locString(), flags=flags)
+        if additionalData.exists("flags"):
+            kwds = dict(flags=additionalData.getInt("flags"))
+        else:
+            kwds = {}
```
parejkoj: Do we want another `if` here for mode="a"? I don't know if there's any way to include "appending" (instead of overwriting) in a butler call.

python/lsst/daf/persistence/posixStorage.py
```diff
@@ -686,10 +686,12 @@ def readFitsCatalogStorage(butlerLocation):
         logLoc = LogicalLocation(locStringWithRoot, additionalData)
         if not os.path.exists(logLoc.locString()):
             raise RuntimeError("No such FITS catalog file: " + logLoc.locString())
-        INT_MIN = -(1 << 31)
-        hdu = additionalData.getInt("hdu", INT_MIN)
-        flags = additionalData.getInt("flags", 0)
-        finalItem = pythonType.readFits(logLoc.locString(), hdu, flags)
+        kwds = {}
+        if additionalData.exists("hdu"):
+            kwds["hdu"] = additionalData.getInt("hdu")
```
parejkoj: This does lose the `INT_MIN ` default value for "hdu", which we should probably keep? There was a ticket on that topic a while back, and I don't recall the results of it...

n8pease: https://github.com/lsst/daf_persistence/commit/ce7e05b8f662d1da2704525e1475d7cd82cc3f55

n8pease: In slack you pointed out the breaking commit, it was https://github.com/lsst/daf_persistence/commit/a76108bce0038784946801f592e244ea7bc19bcf#diff-9c0ba521c98745f8b624916b947c6f4dR698. 
The new code here is actually the old code from that diff, at line 375.
That code was introduced by @TallJimbo, in commit https://github.com/lsst/daf_persistence/commit/ce7e05b8f662d1da2704525e1475d7cd82cc3f55, which was created after I created the ticket branch for my commit a76108bc, and the rebase dropped his change.
So as far as losing INT_MIN goes, you'll have to talk to Jim about that. But as far as correcting my factoring, losing the INT_MIN is the right thing to do.

python/lsst/daf/persistence/posixStorage.py
n8pease: so the same thing here: IMO the fix in this ticket should only be a correct factoring of Jim's commit and if you want to improve on the changes he made I'd like to address that in a separate ticket.

python/lsst/daf/persistence/posixStorage.py
TallJimbo: Yes, we need to drop the `hdu` argument that defaults to `INT_MIN`.  Not all `readFits` methods take that argument (including the one on `PhotoCalib`), so we should only forward it if it's present.

python/lsst/daf/persistence/posixStorage.py
TallJimbo: I don't think we never want the butler to modify a dataset in-place.  If we did, we'd need to first figure  out how that interacts with provenance.

PR #80, opened 2017-09-14
DM-11885: look in daf_persistence for python2-style calls to super() and replace with python 3 style calls.

n8pease:


timj:
I think you need to add
```python
from builtins import super
```
to each file to allow this to work on python2?

n8pease:
I've added `from builtins import super`, but accidentally pushed most of it straight to master: I had done `rebuild lsst_apps` in my py2 env so I could test it and forgot that it checked out master to do that and failed to notice that daf_persistence was no longer on the ticket branch. The commit is https://github.com/lsst/daf_persistence/commit/ee03f118df4c7162fb1368c7c186b059c3213d9b

timj:
Can you fix the commit message of 01765e6 so it doesn't trail off?

n8pease:
oh, hah. weird. yup, will do.

PR #84, opened 2017-11-15
DM-12117 check for old butler parents that are not nested

n8pease:
we started nesting Old Butler parent RepositoryCfgs to preserve
implicit arguments like what calib was used to load them. Some
repositoryCfg.yaml files exist where the Old Butler parent
repository is indicated only by path. This fix internally
converts that path to a RepositoryCfg and the root repository's
mapperArgs are also applied to the internal RepositoryCfg.

### DIFF
--- a/python/lsst/daf/persistence/butler.py
+++ b/python/lsst/daf/persistence/butler.py
@@ -791,6 +791,28 @@ class Butler(object):
                                 root=repoData.repoArgs.cfgRoot,
                                 isV1Repository=isOldButlerRepository)
             else:
+
+                # This is a hack fix for an issue introduced by DM-11284; Old Butler parent repositories used
+                # to be stored as a path to the repository in the parents list and it was changed so that the
+                # whole RepositoryCfg, that described the Old Butler repository (including the mapperArgs that
+                # were used with it), was recorded as a "nested" repository cfg. That checkin did not account
+                # for the fact that there were repositoryCfg.yaml files in the world with only the path to
+                # Old Butler repositories in the parents list.
+                if cfg.parents:
+                    for i, parent in enumerate(cfg.parents):
+                        if isinstance(parent, RepositoryCfg):
+                            continue
+                        parentCfg, parentIsOldButlerRepository = self._getRepositoryCfg(parent)
+                        if parentIsOldButlerRepository:
+                            parentCfg.mapperArgs = cfg.mapperArgs
+                            self.log.info(("Butler is replacing an Old Butler parent repository path '{}' "
+                                           "found in the parents list of a New Butler repositoryCfg: {} "
+                                           "with a repositoryCfg that includes the child repository's "
+                                           "mapperArgs: {}. This affects the instantiated RepositoryCfg "
+                                           "but does not change the persisted child repositoryCfg.yaml file."
+                                           ).format(parent, cfg, parentCfg))
+                            cfg._parents[i] = cfg._normalizeParents(cfg.root, [parentCfg])[0]
+
                 if 'w' in repoData.repoArgs.mode:
                     # if it's an output repository, the RepositoryArgs must match the existing cfg.
                     if not cfgMatchesArgs(repoData.repoArgs, cfg):
--- a/tests/test_DM-12117.py
+++ b/tests/test_DM-12117.py
@@ -1,0 +1,122 @@
+# -*- coding: UTF-8 -*-
+
+#
+# LSST Data Management System
+# Copyright 2016 LSST Corporation.
+#
+# This product includes software developed by the
+# LSST Project (http://www.lsst.org/).
+#
+# This program is free software: you can redistribute it and/or modify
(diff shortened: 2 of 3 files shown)


python/lsst/daf/persistence/butler.py
```diff
@@ -791,6 +791,28 @@ def cfgMatchesArgs(args, cfg):
                                 root=repoData.repoArgs.cfgRoot,
                                 isV1Repository=isOldButlerRepository)
             else:
+
+                # This is a hack fix for an issue introduced by DM-11284; Old Butler parent repositories used
+                # to be stored as a path to the repository in the parents list and it was changed so that the
+                # whole RepositoryCfg, that described the Old Butler repository (including the mapperArgs that
+                # were used with it), was recorded as a "nested" repository cfg. That checkin did not account
+                # for the fact that there were repositoryCfg.yaml files in the world with only the path to
+                # Old Butler repositories in the parents list.
```
parejkoj: Thanks for adding a descriptive note about this block.

tests/test_DM-12117.py
```diff
@@ -0,0 +1,106 @@
...
+_mapper: 'lsst.daf.persistence.test.EmptyTestMapper'
+_mapperArgs: {}
+_parents: ['../repoA']
+_policy: null
+_root: null
+dirty: true
+""".format(mapperArgs)
+
+    def test(self):
+        """Test that an Old Butler parent repo that is can be loaded by a New Butler output repo and that the
```
parejkoj: "that is can be"?

Also, docstrings should be 79 char.

python/lsst/daf/persistence/butler.py
```diff
@@ -1567,6 +1589,29 @@ def _resolveDatasetTypeAlias(self, datasetType):
...
+        if cfg.parents:
+            for i, parent in enumerate(cfg.parents):
+                if isinstance(parent, RepositoryCfg):
+                    continue
+                parentCfg, isOldButlerRepository = self._getRepositoryCfg(parent)
+                if isOldButlerRepository:
+                    cfg._parents[i] = cfg._normalizeParents(cfg.root, [parentCfg])[0]
+
+
+def checkCfgs(cfgList):
```
parejkoj: It's not clear to me where this method gets used: it's not used in this commit anywhere. Also, it has no docstring.

tests/test_DM-12117.py
```diff
@@ -0,0 +1,106 @@
...
+        return """!RepositoryCfg_v1
+_mapper: 'lsst.daf.persistence.test.EmptyTestMapper'
+_mapperArgs: {}
+_parents: ['../repoA']
+_policy: null
+_root: null
+dirty: true
+""".format(mapperArgs)
+
+    def test(self):
```
parejkoj: More descriptive method name than `test`: maybe `test_oldButlerParents` or something?

tests/test_DM-12117.py
```diff
@@ -0,0 +1,106 @@
...
+
+        1. create an Old Butler repo
+        2. create a New Butler repo without mapper args
+        3. reload that New Butler repo without naming its parent as an input
+        4. verify that the parent is loaded as an input
+
+        5. do the same, but specify a mapper arg, and verify that that mapper arg is passed to the parent as
+           well as the root repo.
+        """
+        for mapperArgs in ({}, {'calib': 'foo'}):
```
parejkoj: Doing this as a for loop makes it hard to parse what's going on. I'd suggest writing a helper method `_testSomething/checkSomething` that takes mapperArgs and does what is inside the loop, and then unrolling the loop like this (if I'm interpreting it correctly):

```python
def testSomething(self):
_testSomething({})
_testSomething({'calib':'foo'})
```

This will make test failures much easier to debug, since you won't have to figure out which loop is failing.

Actually, since you're calling `tearDown` manually at the bottom of this, I'd recommend just making each of the above calls its own test method: `testSomethingDefaults` and `testSomethingCalib`. Then `tearDown` will behave as expected. And the method docstring here can go up to the class level.

tests/test_DM-12117.py
```diff
@@ -0,0 +1,106 @@
...
+                f.write(self.repoBYaml(mapperArgs))
+            butler = dp.Butler(repoBPath)
+            self.assertEqual(butler._repos.inputs()[0].repo._mapper.root, repoBPath)
+            self.assertEqual(butler._repos.inputs()[1].repo._mapper.root, repoAPath)
+            self.assertEqual(butler._repos.outputs()[0].repo._mapper.root, repoBPath)
+            self.assertEqual(butler._repos.inputs()[0].repo._mapper.kwargs.get('calib'),
+                             mapperArgs.get('calib'))
+            self.assertEqual(butler._repos.inputs()[1].repo._mapper.kwargs.get('calib'),
+                             mapperArgs.get('calib'))
+            self.tearDown()
```
parejkoj: Don't call `tearDown` manually.

PR #86, opened 2018-03-12
DM-10319: _filename functions return first mappable repository even when file does not exist

erykoff:
As discussed on JIRA, this implementation sidesteps the _filename magic entirely and adds a new getUri(write=write) method for debugging that retrieves the actual filename that the butler will use.  This is what developers expect.

### DIFF
--- a/tests/test_repository.py
+++ b/tests/test_repository.py
@@ -10,7 +10,7 @@
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
 # the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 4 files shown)


tests/test_repository.py
```diff
@@ -10,7 +10,7 @@
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
 # the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
+# (at your option) any later version.1n
```
n8pease: I think this `1n` is a typo?

PR #87, opened 2018-03-14
Catch only NoResults and IOError exceptions.

ktlim:
Bypass function handling was trapping all exceptions, but most of them are actually errors that should be reported.

### DIFF
--- a/python/lsst/daf/persistence/butler.py
+++ b/python/lsst/daf/persistence/butler.py
@@ -1350,14 +1350,14 @@ class Butler(object):
                     # exists in storage, because the bypass function may not actually use the location
                     # according to the template. Instead, execute the bypass function and include its results
                     # in the bypass attribute of the location. The bypass function may fail for any reason,
-                    # the most common case being that a file does not exist. If it raises an exception we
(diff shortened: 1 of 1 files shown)


python/lsst/daf/persistence/butler.py
```diff
@@ -1350,14 +1350,14 @@ def _locate(self, datasetType, dataId, write):
...
-                    # ignore its existance and proceed as though it does not exist.
+                    # the most common case being that a file does not exist. If it raises an exception
+                    # indicating such, we ignore the bypass function and proceed as though it does not exist.
                     if hasattr(location.mapper, "bypass_" + location.datasetType):
                         bypass = self._getBypassFunc(location, dataId)
                         try:
                             bypass = bypass()
                             location.bypass = bypass
-                        except Exception:
+                        except (NoResults, IOError):
```
n8pease: Maybe it would be helpful to log something here? "a bypass function exists but no file was found at {location}" ?? 
I don't know/up to you, I've just been seeing a lot of asks for helpful error messages in recent threads. OTOH it might be noisy in the logs...

PR #90, opened 2018-04-16
DM-13876: Implement and register ParquetStorage as a butler storage type

timothydmorton:
This defines readParquetStorage and writeParquetStorage to define
butler get/put operations for ParquetStorage datatypes.  They use
the wrapper object ParquetTable around pyarrow read/write operations.
Read returns a lazy object; meaning no data is loaded, but arbitrary
columns can be loaded later using the .to_df method of the ParquetTable.

### DIFF
--- a/python/lsst/daf/persistence/posixStorage.py
+++ b/python/lsst/daf/persistence/posixStorage.py
@@ -617,8 +617,8 @@ def readParquetStorage(butlerLocation):
 
     The object returned by this is expected to be a subtype
     of `ParquetTable`, which is a thin wrapper to `pyarrow.ParquetFile`
-    that allows for lazy loading of the data.  This is notably 
-    a different object from what gets passed to the butler `put` function, 
+    that allows for lazy loading of the data.  This is notably
+    a different object from what gets passed to the butler `put` function,
     which should be a `pandas.DataFrame`.
 
     Parameters
--- a/python/lsst/daf/persistence/posixStorage.py
+++ b/python/lsst/daf/persistence/posixStorage.py
@@ -288,7 +288,7 @@ class PosixStorage(StorageInterface):
         raise(RuntimeError("No formatter for location:{}".format(butlerLocation)))
 
     def butlerLocationExists(self, location):
-        """Implementaion of PosixStorage.exists for ButlerLocation objects.
+        """Implementation of PosixStorage.exists for ButlerLocation objects.
         """
         storageName = location.getStorageName()
         if storageName not in ('BoostStorage', 'FitsStorage', 'PafStorage',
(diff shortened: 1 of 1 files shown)


python/lsst/daf/persistence/posixStorage.py
```diff
@@ -610,6 +612,70 @@ def writeFitsStorage(butlerLocation, obj):
             persistence.persist(obj, storageList, butlerLocation.getAdditionalData())
 
 
+def readParquetStorage(butlerLocation):
+    """Read from a butlerLocation.
+
+    The object returned by this is expected to be a subtype
+    of `ParquetTable`, which is a thin wrapper to `pyarrow.ParquetFile`
+    that allows for lazy loading of the data.  This is notably
```
ktlim: This comment is incorrect now, isn't it?

python/lsst/daf/persistence/posixStorage.py
```diff
@@ -610,6 +612,70 @@ def writeFitsStorage(butlerLocation, obj):
...
+        # pythonType will be ParquetTable (or perhaps MultilevelParquetTable)
+        #  filename should be the first kwarg, but being explicit here.
+        results.append(pythonType(filename=filename))
+    return results
+
+
+def writeParquetStorage(butlerLocation, obj):
+    """Writes pandas dataframe to parquet file
+
+    Note, this takes a `pandas.DataFrame` to write, whereas
```
ktlim: And this comment; I think it expects a `ParquetTable` as well.

python/lsst/daf/persistence/posixStorage.py
```diff
@@ -610,6 +612,70 @@ def writeFitsStorage(butlerLocation, obj):
...
+    which should be a `pandas.DataFrame`.
+
+    Parameters
+    ----------
+    butlerLocation : ButlerLocation
+        The location & formatting for the object(s) to be read.
+
+    Returns
+    -------
+    A list of objects as described by the butler location. One item for
```
ktlim: The original purpose of the list of locations was to allow a single dataset to be assembled from multiple physical serializations.  If you're not actually expecting/using the list, I would just use the initial entry (and perhaps even warn if there is more than one).

python/lsst/daf/persistence/posixStorage.py
```diff
@@ -288,10 +288,12 @@ def read(self, butlerLocation):
         raise(RuntimeError("No formatter for location:{}".format(butlerLocation)))
 
     def butlerLocationExists(self, location):
-        """Implementaion of PosixStorage.exists for ButlerLocation objects."""
+        """Implementaion of PosixStorage.exists for ButlerLocation objects.
```
ktlim: Could you fix the typo here while you're in here?

PR #92, opened 2018-05-08
DM-14356: Add matplotlib.figure.Figure storage to PosixStorage.

TallJimbo:
Matplotlib figures cannot be retrieved using the butler, but 'put' should now work.

Note that this storage actually only requires that the object being saved have a `savefig` method with the same behavior as matplotlib's.

### DIFF
--- a/python/lsst/daf/persistence/posixStorage.py
+++ b/python/lsst/daf/persistence/posixStorage.py
@@ -777,7 +777,6 @@ def writeFitsCatalogStorage(butlerLocation, obj):
         else:
             kwds = {}
         obj.writeFits(logLoc.locString(), **kwds)
-        return
 
 
 def readMatplotlibStorage(butlerLocation):
--- a/python/lsst/daf/persistence/posixStorage.py
+++ b/python/lsst/daf/persistence/posixStorage.py
@@ -293,7 +293,7 @@ class PosixStorage(StorageInterface):
         storageName = location.getStorageName()
         if storageName not in ('BoostStorage', 'FitsStorage', 'PafStorage',
                                'PickleStorage', 'ConfigStorage', 'FitsCatalogStorage',
-                               'ParquetStorage'):
+                               'ParquetStorage', 'MatplotlibStorage'):
             self.log.warn("butlerLocationExists for non-supported storage %s" % location)
             return False
         for locationString in location.getLocations():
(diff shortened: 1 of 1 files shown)


python/lsst/daf/persistence/posixStorage.py
```diff
@@ -780,6 +780,57 @@ def writeFitsCatalogStorage(butlerLocation, obj):
...
+    ----------
+    butlerLocation : ButlerLocation
+        The location & formatting for the object to be written.
+    obj : matplotlib.figure.Figure
+        The object to be written.
+
+    Raises
+    ------
+    ValueError
+        Raised if the file
```
laurenam: Trailing sentence.  I also don't see a **raise** below.

python/lsst/daf/persistence/posixStorage.py
```diff
@@ -780,6 +780,57 @@ def writeFitsCatalogStorage(butlerLocation, obj):
...
+        # and pass that as the format directly.
+        _, ext = os.path.splitext(locations[0])
+        if ext:
+            ext = ext[1:]  # strip off leading '.'
+        else:
+            # If there is no extension, we let matplotlib fall back to its
+            # default.
+            ext = None
+        obj.savefig(logLoc.locString(), format=ext)
+        return
```
laurenam: Is the **return** meant to be indented to within the **with SafeFilename** here?

PaulPrice: The empty `return` here serves no purpose, and should be removed.

python/lsst/daf/persistence/posixStorage.py
```diff
@@ -780,6 +780,57 @@ def writeFitsCatalogStorage(butlerLocation, obj):
...
+    ValueError
+        Raised if the file
+    """
+    additionalData = butlerLocation.getAdditionalData()
+    locations = butlerLocation.getLocations()
+    with SafeFilename(os.path.join(butlerLocation.getStorage().root, locations[0])) as locationString:
+        logLoc = LogicalLocation(locationString, additionalData)
+        # SafeFilename appends a random suffix, which corrupts the extension
+        # matplotlib uses to guess the file format.
+        # Instead, we strip off the extension from the original location
```
wmwv: I would say "extract the extension".
"strip off the extension" implies it's removed from the location name.

PR #93, opened 2018-05-11
DM-14359: Fix data ID handling in ap_*

kfindeisen:
This PR factors some of the dataId handling logic out of `pipe_base` so that it can be used by code other than `ArgumentParser`. I've created a new module, `butlerHelpers.py`, for similar application logic that may be factored out in the future (in the unlikely event that it's needed before we rewrite everything for Butler 3).

### DIFF
--- a/python/lsst/daf/persistence/butlerHelpers.py
+++ b/python/lsst/daf/persistence/butlerHelpers.py
@@ -1,0 +1,80 @@
+#
+# LSST Data Management System
+# Copyright 2008-2018 AURA/LSST.
+#
+# This product includes software developed by the
+# LSST Project (http://www.lsst.org/).
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+
+"""Common tasks and idioms performed with the Butler.
+"""
+
+__all__ = ["dataExists", "searchDataRefs"]
+
+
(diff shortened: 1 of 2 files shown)


python/lsst/daf/persistence/butlerHelpers.py
```diff
@@ -0,0 +1,80 @@
...
+
+    Returns
+    -------
+    exists : `bool`
+        Return value is `True` if data exists, `False` otherwise.
+    """
+    subDRList = dataRef.subItems()
+    if subDRList:
+        for subDR in subDRList:
+            if dataExists(subDR):
```
parejkoj: I am curious why this needs to be recursive. Wouldn't just `datasetExists` handlle all such situations? When does it not?

python/lsst/daf/persistence/butlerHelpers.py
```diff
@@ -0,0 +1,80 @@
...
+
+    Parameters
+    ----------
+    butler: `lsst.daf.persistence.Butler`
+        The repository to query for data.
+    datasetType : `str`
+        The type of data references to return.
+    level : `str`
+        The level of data ID at which to search. If the empty string, the
+        default level for ``datasetType`` shall be used.
```
parejkoj: Please reference wherever it is in the butler docs that `level` is better defined. If anywhere.

python/lsst/daf/persistence/butlerHelpers.py
kfindeisen: I have no idea. This is just copied from `pipe_base`.

python/lsst/daf/persistence/butlerHelpers.py
kfindeisen: I'm not aware of any definition of `level` in the docs.

python/lsst/daf/persistence/butlerHelpers.py
kfindeisen: It looks like @r-owen wrote this function [originally](https://github.com/lsst/pipe_base/commit/792b4b912d5fcb80de9d88a22737de01c88f61e3).

python/lsst/daf/persistence/butlerHelpers.py
r-owen: I suggest we leave it alone. It was necessary in the butler when I wrote it, because of dataset types such as raw images, which were per amplifier. I strongly suspect it's still necessary. It will go away with gen3 butler, thank heavens. I hated the code when I wrote it and I still hate it.

PR #98, opened 2018-08-30
DM-15599: Stop using boost persistence in afw

r-owen:


python/lsst/daf/persistence/posixStorage.py
```diff
@@ -561,7 +562,74 @@ def writeConfigStorage(butlerLocation, obj):
...
+    with SafeFilename(os.path.join(butlerLocation.getStorage().root, locations[0])) as locationString:
+        logLoc = LogicalLocation(locationString, additionalData)
+        if supportsOptions:
+            obj.writeFitsWithOptions(logLoc.locString(), options=additionalData)
+        else:
+            obj.writeFits(logLoc.locString())
+
+
+def boostReadFitsStorage(butlerLocation):
+    """Read a FITS image from a butlerLocation using boost peristence
```
TallJimbo: Seems like we should remove this function and check that we don't actually use it anywhere anymore.

I also don't think it actually makes sense - we've never used Boost to read FITS code.  At most this could be using Boost to read `daf::base::Persistable`, but that has nothing to do with FITS.

r-owen: Sounds great! I will try that and see what Jenkins thinks.

r-owen: You were right. After removing this Jenkins still ran, including ci_hsc! Please have another look.

PR #100, opened 2018-09-17
DM-15776: Reimplement FitsStorage support for reading headers directly.

TallJimbo:


TallJimbo:
Yeah, I think that would require (at least) moving the test for what kind of reader to use inside the loop over locations so we could capture the concrete options in closure (at least I think that's what you had in mind).  While that loop is only ever going to have one iteration, that was enough to make that option seem to me to be not obviously better than what I have there now.

PR #101, opened 2018-09-28
Workaround for ignoring the requested HDU; DM-15902

RobertLuptonTheGood:


python/lsst/daf/persistence/posixStorage.py
```diff
@@ -604,7 +604,16 @@ def readFitsStorage(butlerLocation):
         if supportsOptions:
             finalItem = pythonType.readFitsWithOptions(logLoc.locString(), options=additionalData)
         else:
-            finalItem = reader(logLoc.locString())
+            fileName = logLoc.locString()
+            mat = re.search(r"^(.*)\[(\d+)\]$", fileName)
+
+            if mat and reader == readMetadata: # readMetadata() only understands the hdu argument, not [hdu]
```
TallJimbo: Instead of just `reader == readMetadata`, it might be good to also include the explicit HDU arg if `issubclass(pythonType, (afwImage.image, afwImage.Mask)`, because we may want to read specific HDUs according to the template for those, too.  It looks like neither `MaskedImage` nor `Exposure` have an `hdu` argument to `readFits` (which makes sense).

It may also be a good idea to reorder the logic, and only to the regex/stripping if the reader is [one of] the appropriate type[s].

PR #104, opened 2018-10-26
DM-12390: Butler doesn't raise when failing to write data

n8pease:


n8pease:
@erykoff `getLocations` docs say 
```
        Returns
        -------
        If write is False, will return either a single object or None. If write is True, will return a list
        (which may be empty)
```
it looks like `getUri` works because in the case where `write` is `False` it returns an object or `None` and so the return at https://github.com/lsst/daf_persistence/blob/master/python/lsst/daf/persistence/butler.py#L1536 works.

And in the case where `write` is `True` it iterates over locations, which is an empty list in our case, and ends up raising because it didn't return, here: https://github.com/lsst/daf_persistence/blob/master/python/lsst/daf/persistence/butler.py#L1548

It would probably be safer to change the check to `not locations`, I'll leave that to you decide and do if you want.

n8pease:
`NoResults` seems like a reasonable exception to throw, I'll make that change.

PR #110, opened 2019-03-14
DM-18488: Support pyyaml 5.1

timj:


python/lsst/daf/persistence/butlerLocation.py
```diff
@@ -193,7 +193,11 @@ class ButlerLocation(yaml.YAMLObject):
     """
 
     yaml_tag = u"!ButlerLocation"
-    yaml_loader = yaml.Loader
+    try:
```
ktlim: Is this actually used anywhere?  It seems strange that this is a class variable rather than built into `from_yaml`, and I don't see any caller for the latter or use of the variable.

python/lsst/daf/persistence/repositoryCfg.py
```diff
@@ -317,4 +317,15 @@ def __repr__(self):
             self._policy)
 
 
-yaml.add_constructor(u"!RepositoryCfg_v1", RepositoryCfg.v1Constructor)
+Loaders = [yaml.Loader, ]
```
ktlim: Maybe `loaderList` rather than `Loaders`?  Otherwise, it looks like a class name.

python/lsst/daf/persistence/butlerLocation.py
timj: Weird, I changed it because I assumed it was used but you are right, I can't find anywhere that uses that or the dumper.

timj: My mistake. This class inherits from `yaml.YamlObject` and `yaml_loader` is the class variable that `yaml` assumes will define the loader.

PR #111, opened 2019-04-26
DM-12594: Remove daf::base::Citizen

parejkoj:


src/DbAuth.cc
```diff
@@ -62,97 +60,79 @@ namespace pexPolicy = lsst::pex::policy;
...
         if (ret != 0 || (st.st_mode & (S_IRWXG | S_IRWXO)) != 0) {
-            throw LSST_EXCEPT(pexExcept::RuntimeError,
-                    filename + " is missing or accessible by others");
-        }
-        {
-            dafBase::PersistentCitizenScope scopeGuard;
-            authPolicy = pexPolicy::Policy::Ptr(new pexPolicy::Policy(filename));
+            throw LSST_EXCEPT(pexExcept::RuntimeError, filename + " is missing or accessible by others");
         }
+        { authPolicy = pexPolicy::Policy::Ptr(new pexPolicy::Policy(filename)); }
```
natelust: Why is this happening in its own execution block?

parejkoj: Ha! Look at the diff above: this was just collapsed by clang-format from lines 89-94 in the original.

So, I have no idea why it's a separate block.

natelust: I think I vaguely recall either Jim or Pim working on something to to with the scopeGaurd. I think it was related to a ticket Pim had worked on to make daf persistence thread safe, which is what the scope had to be in place. Now that that is removed I think the scope block can be removed too.

parejkoj: You may be right about the reasoning, but according to git history, the `scopeGuard` was added 10 years ago by `smm`.

https://github.com/lsst/daf_persistence/commit/58860a4a4f0a7d5db62405f16e1edc305eb56e23

I'm tempted to just leave it as is, given the imminent removal of `daf_persistence`. ;-)

PR #121, opened 2020-10-31
DM-22814: Remove afw APIs deprecated in DM-17566

parejkoj:


python/lsst/daf/persistence/posixStorage.py
```diff
@@ -580,7 +580,7 @@ def readFitsStorage(butlerLocation):
     if not supportsOptions:
         from lsst.daf.base import PropertySet, PropertyList
         if issubclass(pythonType, (PropertySet, PropertyList)):
-            from lsst.afw.image import readMetadata
+            from lsst.afw.fits import readMetadata
```
kfindeisen, line 583: Out of scope, but maybe get rid of the local import while you're at it?

kfindeisen (approved):

python/lsst/daf/persistence/posixStorage.py
parejkoj: I think I'll leave it, given daf_persistence near-obsolescence (tm).

PR #125, opened 2021-02-04
DM-28659: Change 1000 to 3 writers in test.

ktlim:


tests/test_safeFileIo.py
```diff
@@ -147,14 +147,14 @@ def testNoChange(self):
...
-class TestOneThousandWriters(unittest.TestCase):
+class TestMultipleWriters(unittest.TestCase):
     """Test for efficient file updating with shared & exclusive locks by serializing a RepositoryCfg to a
     location 1000 times. When this was tested on a 2.8 GHz Intel Core i7 macbook pro it took about 1.3 seconds
     to run. When the squash performance monitoring framework is done, this test could be monitored in that
     system."""
 
     def setUp(self):
-        self.testDir = tempfile.mkdtemp(dir=ROOT, prefix='TestOneThousandWriters-')
+        self.testDir = tempfile.mkdtemp(dir=ROOT, prefix='TestMultipleWriters-')
```
parejkoj, line 157: Why not do `self.__class__.__name__` for the prefix here, to keep it generic?

tests/test_safeFileIo.py
```diff
@@ -147,14 +147,14 @@ def testNoChange(self):
             self.assertEqual(f.read(), "some test string")
 
 
-class TestOneThousandWriters(unittest.TestCase):
+class TestMultipleWriters(unittest.TestCase):
     """Test for efficient file updating with shared & exclusive locks by serializing a RepositoryCfg to a
     location 1000 times. When this was tested on a 2.8 GHz Intel Core i7 macbook pro it took about 1.3 seconds
```
parejkoj, line 152: This is no longer 1000 times. Please update the docstring if anything else needs to change, too.

parejkoj (commented):
Some docstring suggestions.

tests/test_safeFileIo.py
ktlim: Seems a bit too fancy for something that is scheduled to die before the end of the year.

PR #127, opened 2021-03-29
DM-29497: Issue deprecation warning for Mapper/Butler

timj:
This code implements three options:

* Only issues one ever
* Issues one per component (Butler or Mapper)
* Always issues

I've defaulted it to only one ever for now.

### DIFF
--- a/python/lsst/daf/persistence/deprecation.py
+++ b/python/lsst/daf/persistence/deprecation.py
@@ -22,6 +22,7 @@
 
 __all__ = ("deprecateGen2", "always_warn", "deprecate_class")
 
+import textwrap
 import traceback
 import warnings
 
@@ -35,7 +36,15 @@ Options are:
 * `None`: Only issue a warning once regardless of component.
 """
 
+# Cache recording which components have issued previously
 _issued = {}
+
+# This is the warning message to issue. There is a "label" placeholder
+# that should be inserted on format.
+_warning_msg = "Gen2 Butler has been deprecated{label}. It will be removed sometime after v23.0 but no" \
+               " earlier than the end of 2021."
+
+_version_deprecated = "v22.0"
 
 
 def deprecateGen2(component=None):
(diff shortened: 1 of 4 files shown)


python/lsst/daf/persistence/mapper.py
```diff
@@ -96,6 +97,7 @@ def __new__(cls, *args, **kwargs):
         return self
 
     def __init__(self, **kwargs):
+        deprecateGen2("Mapper")
```
ktlim, line 100: Have we checked that all `Mapper` subclasses call `super().__init__`?

python/lsst/daf/persistence/deprecation.py
```diff
@@ -0,0 +1,75 @@
...
+    # and into user code.
+    stacklevel = 3  # Default to the caller's caller
+    stack = traceback.extract_stack()
+    for i, s in enumerate(reversed(stack)):
+        if not ("python/lsst/daf/persistence" in s.filename or "python/lsst/obs/base" in s.filename):
+            stacklevel = i + 1
+            break
+
+    label = ""
+    if component is not None and ALWAYS is not None:
```
ktlim, line 70: Maybe parallel the ifs above and reverse the order of these conditions?

ktlim (approved):

python/lsst/daf/persistence/deprecation.py
```diff
@@ -0,0 +1,75 @@
...
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+
+__all__ = ("deprecateGen2",)
+
+import traceback
+import warnings
+
+ALWAYS = None
```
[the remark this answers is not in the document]
ktlim: Oh, I get it, despite its use in the test, it _is_ supposed to be a hard-coded constant, right?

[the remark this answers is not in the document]
timj: Yes, I'm not sure how we want to play it. I was leaving the ALWAYS=True option for when we need to get serious about removing things. I'm fine with making it a public variable.

python/lsst/daf/persistence/mapper.py
timj: No. To a certain extent instantiating a `Butler` will get most places.

timj: I can add an `__init_subclass__` to `Mapper` to deal with that I think.

timj: Of course `__init_subclass__` doesn't help us because that triggers at import and not object construction. Maybe I have to insert a `__new__`?

timj: Rewrote to use a class decorator.

PR #128, opened 2021-05-26
DM-30004: replace boost::regex with std::regex

mwittgen:


ktlim (approved):
You could use `auto` when constructing the iterator, and you could avoid constructing the termination flag in each loop, but those are minor code optimizations that are not worth doing in code that is deprecated.

PR #130, opened 2021-08-18
DM-31448: Allow gen2 butler lookups even when fields in the dataId are not in the registry

RobertLuptonTheGood:


python/lsst/daf/persistence/registries.py
```diff
@@ -339,23 +344,22 @@ def lookup(self, lookupProperties, reference, dataId, **kwargs):
...
         cmd = "SELECT DISTINCT "
         cmd += ", ".join(lookupProperties)
         cmd += " FROM " + " NATURAL JOIN ".join(reference)
         valueList = []
         if dataId is not None and len(dataId) > 0:
             whereList = []
             for k, v in dataId.items():
+                if checkColumns:        # check if k is in registry
+                    try:
+                        self.conn.cursor().execute(f"select count({k}) from {reference[0]}")
```
ktlim, line 359: This is potentially very slow, and I don't think it's guaranteed that `reference[0]` holds all columns of interest.

As a first step, `SELECT {k} FROM {reference[0]} LIMIT 1` is likely to be much more efficient.

As a next step, `SELECT * FROM {reference[0]} LIMIT 1` and inspecting the returned row's schema would be even better.

Finally, the `NATURAL JOIN`-based `FROM` clause could be used to look at all available tables.

RobertLuptonTheGood: I suppose I was assuming an index on the table so the `count(*)` should be fast, but I'm happy to use the `LIMIT 1`; and I was just being lazy with the `JOIN`.
However, inspecting the schema seems un-necessary (and more work).  What is gained over `f'SELECT {k} FROM {" NATURAL JOIN ".join(reference)} LIMIT 1'`?

ktlim: Issuing one query for `SELECT *` and seeing the returned columns is going to be much more efficient than `SELECT {k}` for each column as long as the number of columns is much greater than 1.  In this case, though, I guess the columns are not all the ones in the database but instead all the ones in the `dataId`.  So maybe you really want `SELECT {", ".join(dataId.keys())} FROM ...`

ktlim: Ah, reading still closer, you want individual column-based notification of presence.  So that gets you back to `SELECT *` if the `dataId` contains a lot of columns or your current code if it only has a few.

python/lsst/daf/persistence/registries.py
```diff
@@ -363,6 +368,37 @@ def lookup(self, lookupProperties, reference, dataId, **kwargs):
...
+        e.g. if the template is 'raw/raw_v%(visit)d_f%(filter)s.fits.gz', and
+        dataId={'visit':1}, and lookupProperties is ['filter'], and the
+        filesystem under self.root has exactly one file 'raw/raw_v1_fg.fits.gz'
+        then the return value will be [('g',)]
+
+        :param lookupProperties:
+        :param dataId: must be an iterable. Keys must be string.
+        If key is a string then will look for elements in the repository that match value for key.
+        If key is a 2-item iterable then will look for elements in the repository where the value is between
+        the values of key[0] and key[1].
```
ktlim, line 385:
```suggestion
        :param dataId: must be a key/value iterable. Keys must be string.
        If value is a string then will look for elements in the repository that match that value.
        If value is a 2-item iterable then will look for elements in the repository where the value is between
        the values of value[0] and value[1].
```

ktlim: Kind of duplicative to have this both here and above; maybe remove one?

python/lsst/daf/persistence/registries.py
```diff
@@ -363,6 +368,37 @@ def lookup(self, lookupProperties, reference, dataId, **kwargs):
...
+        if not self.conn:
+            return None
+
+        # input variable sanitization:
+        reference = sequencify(reference)
+        lookupProperties = sequencify(lookupProperties)
+
+        try:
+            return self._lookup(lookupProperties, dataId, reference)
+        except sqlite3.OperationalError:
```
ktlim, line 399: It's just a performance optimization not to have `checkColumns` always on?

RobertLuptonTheGood: That's right.  I added some text to clarify this.

python/lsst/daf/persistence/registries.py
RobertLuptonTheGood: Done

python/lsst/daf/persistence/registries.py
RobertLuptonTheGood: I think we'll usually have only a few fields, and as I need to know which one caused the problem I think I'll stick with what I have.  As we note below, this code only fires when there are fields in the dataId that confuse the registry query, which is moderately uncommon so a little inefficiency is probably OK.  The use case that forced me to create this patch was specifying an int to read a particular HDU when the dataId was unambiguous but not complete.

RobertLuptonTheGood: Oh, and I accepted the don't-use-count, use limit, and ask for the join suggestions.

PR #132, opened 2022-04-26
DM-34559: Make deprecation warning even more dire

timj:


python/lsst/daf/persistence/deprecation.py
```diff
@@ -42,7 +42,8 @@
 # This is the warning message to issue. There is a "label" placeholder
 # that should be inserted on format.
 _warning_msg = "Gen2 Butler has been deprecated{label}. It will be removed sometime after v23.0 but no" \
-               " earlier than the end of 2021."
+               " earlier than the end of 2021." \
```
ktlim, line 45: I'd say:
```
Gen2 Butler has been deprecated{label}.  This Gen2 code may be removed in any future daily or weekly release.
```
It won't stop working in a given current release.  We shouldn't have it be present but broken.  And I don't think the dates say much at this point.

ktlim (approved):

python/lsst/daf/persistence/deprecation.py
```diff
@@ -41,8 +41,8 @@
 
 # This is the warning message to issue. There is a "label" placeholder
 # that should be inserted on format.
-_warning_msg = "Gen2 Butler has been deprecated{label}. It will be removed sometime after v23.0 but no" \
-               " earlier than the end of 2021."
+_warning_msg = "Gen2 Butler has been deprecated{label}. "\
+               " This Gen2 code may be removed in any future daily or weekly release."
```
mrawls, line 45: This is nicer than the previous message, but still doesn't give any context for what Gen2 means for a user who has just been doing what they thought were Science Pipeline Things.

Perhaps add, "For a Gen3-based Getting Started Tutorial, please see `https://pipelines.lsst.io/getting-started/index.html`." I think linking to something marginally useful now is better than leaving users in the dark by not linking to anything because it's not ready or perfect yet.

timj: Fair enough. I'll make that change.

PR #135, opened 2022-07-26
Support a bypass function to check for the existence of a bypassed datatype

RobertLuptonTheGood:


python/lsst/daf/persistence/butler.py
```diff
@@ -1366,12 +1377,15 @@ def _locate(self, datasetType, dataId, write):
         return locations
 
     @staticmethod
-    def _getBypassFunc(location, dataId):
+    def _getBypassFunc(location, dataId, bypassBasename="bypass_"):
```
PaulPrice, line 1380: Needs docstring.

python/lsst/daf/persistence/butler.py
```diff
@@ -1337,15 +1340,23 @@ def _locate(self, datasetType, dataId, write):
...
+                    if checkExistence:
+                        bypass = self._getBypassFunc(location, dataId, "bypass_datasetExists_")
+                    if not bypass:
                         bypass = self._getBypassFunc(location, dataId)
+
+                    if bypass:
                         try:
-                            bypass = bypass()
-                            location.bypass = bypass
+                            location.bypass = bypass()
```
PaulPrice, line 1359: So sometimes `location.bypass` is a `bool` indicating mere existence, and sometimes it's a heavy object like `Exposure`? That doesn't feel right. I think the logic needs cleaning up. Maybe just a `bypass() is None`, but then that completely changes the meaning of `location.bypass`. Maybe some thought needs to go into this, or maybe it doesn't given that this is now effectively a dead package?

python/lsst/daf/persistence/butler.py
RobertLuptonTheGood: It's not a new function, but OK

python/lsst/daf/persistence/butler.py
RobertLuptonTheGood: The code passed `location.bypass` around, so it's much simpler to use it as is.  And the code is dead, so I'd rather keep it this way.  Twist my arm if needs be

